PR #1, opened 2023-11-29
Allow multiple CHEFS forms to be configured/displayed

kyle1morel:
Allow multiple CHEFS forms to be configured on the back end. Requests to the CHEFS API from the back end will create an Axios request with a dynamic basic auth by doing a lookup into the config map based on the form ID in the request. Submission data will be merged together for every form that is configured, before finally being filtered by user (if required) and sent to the front end.

# Description

## Types of changes

## Checklist


wilwong89 (approved):
Provisionally approving, needs to fix failing tests.

app/src/controllers/chefs.ts
```diff
@@ -1,84 +1,55 @@
...
+        })
+      );
+
+      /*
+       * Filter Data source
+       * IDIR users should be able to see all submissions
+       * BCeID/Business should only see their own submissions
+       */
+      const filterData = (data: Array<ChefsSubmissionDataSource>) => {
+        const filterToUser = (req.currentUser?.tokenPayload as JwtPayload).identity_provider !== IdentityProvider.IDIR;
```
jujaga, line 31: Just as a note - in the event we use a non-SSO Keycloak IAM we might not necessarily have the `identity_provider` attribute available. This is one of those few situations where I'm thinking it may be worth doing a defensive null check for tokenPayload?

app/src/services/chefs.ts
```diff
@@ -2,92 +2,38 @@
...
  */
-function chefsAxios(options: AxiosRequestConfig = {}): AxiosInstance {
-  const instance = axios.create({
+function chefsAxios(formId: string, options: AxiosRequestConfig = {}): AxiosInstance {
+  return axios.create({
     baseURL: config.get('frontend.chefs.apiPath'),
     timeout: 10000,
-    ...options
+    ...options,
+    auth: { username: formId, password: getChefsApiKey(formId) ?? '' }
```
jujaga, line 20: May be better to have the auth attribute declared before the final ...options - just in the event you want to directly override the auth value via AxiosRequestConfig in the service layer you would be able to do so - ordered the way it is right now it's not configurable.

app/src/types/ChefsFormConfig.ts
```diff
@@ -0,0 +1,10 @@
+export type ChefsFormConfig = {
+  form1: ChefsFormConfigData;
+  form2: ChefsFormConfigData;
+};
+
+export type ChefsFormConfigData = {
+  name: string;
+  formId: string;
+  formApiKey: string;
```
jujaga, line 9: Could be 'id' and 'apiKey' from earlier style comment if we wanted to be more terse.

frontend/src/router/index.ts
```diff
@@ -23,6 +23,12 @@ const routes: Array<RouteRecordRaw> = [
     component: () => import('../views/HomeView.vue'),
     meta: { title: 'Home' }
   },
+  {
+    path: '/initiatives',
+    name: RouteNames.INITIATIVES,
+    component: () => import('../views/InitiativesView.vue'),
+    meta: { title: 'Initiatives' }
```
jujaga, line 30: If this is to be an authentication-gated page, we might need that authentication thingie we usually do here?

jujaga (commented):
Seems good for the most part - just a few minor odds and ends to address, including the failed linting/tests.

app/src/components/utils.ts
```diff
@@ -1,9 +1,21 @@
+import config from 'config';
 import { existsSync, readFileSync } from 'fs';
 import { join } from 'path';
 
 import { getLogger } from './log';
+import { ChefsFormConfig, ChefsFormConfigData } from '../types/ChefsFormConfig';
```
TimCsaky, line 6: this may sound like a lazy excuse.. but i would be tempted to just use more highlevel types like `object`, `array`, etc. where its safe to do that. especially while this app and the data model is still evolving

app/src/components/utils.ts
```diff
@@ -1,9 +1,21 @@
...
 
+/**
+ * @function getChefsApiKey
+ * Search for a CHEFS form Api Key
+ * @returns {string | undefined} The CHEFS form Api Key if it exists
+ */
+export function getChefsApiKey(formId: string): string | undefined {
+  const cfg = config.get('server.chefs.forms') as ChefsFormConfig;
+  return Object.values<ChefsFormConfigData>(cfg).find((o: ChefsFormConfigData) => o.formId === formId)?.formApiKey;
+}
```
TimCsaky, line 17: nice use of a function here

TimCsaky (approved):
This is looking great!

jujaga (approved):

jatindersingh93 (approved):

PR #2, opened 2023-11-30
Full data display in CHEFS Submissions table

wilwong89:
# Description
Loads all necessary table data via the CHEFS export endpoint.
Updates PrimeVue to 3.42.0. Implements CSS changes to standardize with other Common Services applications.

## Types of changes

## Checklist


kyle1morel (changes requested):
The approach is good with lazy loading the true submission data as needed, but the current implementation completely falls apart when you begin sorting. This will need to be addressed before merging. Recommend tracking the values coming back from the DataTable @page and @value-change emits. Use those values with the @update:sort-field and @update:sort-order callbacks to correctly get the current displayed rows to load in.

kyle1morel (approved):

jujaga (approved):

app/src/routes/v1/chefs.ts
```diff
@@ -9,8 +9,8 @@ const router = express.Router();
 router.use(requireSomeAuth);
 
 // Submission endpoint
-router.get('/submissions', (req: Request, res: Response, next: NextFunction): void => {
-  chefsController.getSubmissions(req, res, next);
+router.get('/export', (req: Request, res: Response, next: NextFunction): void => {
+  chefsController.getFormExport(req, res, next);
```
TimCsaky, line 13: minor point: maybe at some point, consider calling this `submissionExport`.  form export is another chefs feature. i remember from working with chefs it was useful to differentiate between submissions and the form itself.

app/src/services/chefs.ts
```diff
@@ -22,9 +22,11 @@ function chefsAxios(formId: string, options: AxiosRequestConfig = {}): AxiosInst
 }
 
 const service = {
-  getFormSubmissions: async (formId: string) => {
+  getFormExport: async (formId: string) => {
     try {
-      const response = await chefsAxios(formId).get(`forms/${formId}/submissions`);
+      const response = await chefsAxios(formId).get(`forms/${formId}/export`, {
```
TimCsaky, line 27: although the path itself uses form entity 🤷

PR #4, opened 2023-12-12
Prisma DB connection & Submission Form

kyle1morel:
# Description

Two pronged PR. 

Implements a new DB connection via Prisma. We have thrown Objection.js out the window due to it being sunset in Nov '22. Knex remains for migrations as it allows easy up/down migrations. Prisma is a forward only model which we didn't want. 

Also implements the submission form in a read-only and editable format. Data is E2E. So it is being read and saved from our database.

## Types of changes

## Checklist


app/src/db/migrations/20231212000000_init.ts
```diff
@@ -0,0 +1,176 @@
...
+          table.string('firstName', 255);
+          table.string('fullName', 255);
+          table.string('lastName', 255);
+          table.boolean('active').notNullable().defaultTo(true);
+          stamps(knex, table);
+        })
+      )
+
+      .then(() =>
+        knex.schema.createTable('submission', (table) => {
```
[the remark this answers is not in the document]
TimCsaky: also with this many columns, i wonder if they should be ordered logically (eg alphabetical).

app/knexfile.ts
```diff
@@ -0,0 +1,66 @@
...
+ * @module knexfile
+ * @see module:knex
+ * @see module:config
+ */
+
+import { types } from 'pg';
+
+// To handle JSON Schema validation, we treat dates and timestamps outside of the database as strings.
+// Prevent the automatic conversion of dates/timestamps into Objects, leave as strings.
+types.setTypeParser(1082, (value) => {
```
TimCsaky, line 25: not sure we need this stuff (?)

TimCsaky (approved):
looking good!

frontend/src/utils/constants.ts
```diff
@@ -36,3 +36,9 @@ export const ToastTimeout = Object.freeze({
   SUCCESS: 3000,
   WARNING: 5000
 });
+
+/**
+ * CHEFS form constants
+ */
+export const ApplicationStatusList = ['IN PROGRESS', 'DELAYED', 'COMPLETED'];
+export const IntakeStatusList = ['SUBMITTED', 'ASSIGNED', 'COMPLETED'];
```
jujaga, line 44: Enum candidate?

app/src/db/migrations/20231212000000_init.ts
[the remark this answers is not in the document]
jujaga: Yes I'd second the value alphabetization notion.

app/src/db/migrations/20231212000000_init.ts
```diff
@@ -41,27 +36,27 @@ export async function up(knex: Knex): Promise<void> {
...
-          table.string('streetAddress', 255);
-          table.string('atsClientNumber', 255);
+          table.text('confirmationId');
+          table.text('contactEmail');
+          table.text('contactPhoneNumber');
+          table.text('contactFirstName');
+          table.text('contactLastName');
+          table.text('intakeStatus');
+          table.text('projectName');
+          table.text('queuePriority');
```
jujaga, line 46: Is queuePriority still a string or is it treated as a number?

app/src/db/migrations/20231212000000_init.ts
```diff
@@ -41,27 +36,27 @@ export async function up(knex: Knex): Promise<void> {
...
+          table.text('singleFamilyUnits');
```
jujaga, line 47: This could feel like a number... though if this is correlated to the housing units column... it's technically behaving as a potential number range. "Should" be fine if we maybe leave as a string, but if we ever need to semantically use this down the line it might be better to consider if we do need to bifurcate this into lower and upper number columns instead...

app/src/components/utils.ts
jujaga, line 1: I think we need to be much stricter with how the toYrn and fromYrn argument typings are to properly leverage Typescript behaviors. Allowing a union of undefined negates a large chunk of the value of having a well-defined YRN type to begin with and forces our utility functions to have to have defensive code to handle border cases that shouldn't ever happen to begin with if strictly typed.

jujaga (approved):

app/src/services/chefs.ts
```diff
@@ -33,10 +38,74 @@ const service = {
...
-      const response = await chefsAxios(formId).get(`submissions/${formSubmissionId}`);
-      return response.data;
+      // Try to pull data from our DB
+      let result = await prisma.submission.findUnique({
+        where: {
+          submissionId: submissionId
+        }
+      });
+
+      // Pull submission data from CHEFS and store to our DB if it doesn't exist
```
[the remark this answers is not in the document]
wilwong89: Didn't we come to the agreement that once intake is done with CHEFS, we completely ignore it after? Part of a first step of eventually moving off CHEFS. And the source of truth would be this DB.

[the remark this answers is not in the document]
TimCsaky: yes. ignore my comment. when the navigators start using PCNS we can do a fresh import of submission data from chefs.
but who knows what data they will be adding to the submissions/status/notes attachments etc in the meantime

norrisng-bc (approved):

PR #9, opened 2023-12-20
Assign a user to submission

kyle1morel:
# Description

Implements the User service to automatically create/update users in the database as they log in.
Updates the Helmet content security policy.
Changes the application name across the board to the new one.
Implements assigning of a user to a submission.
Implements a to/from DB model to correctly convert objects from Prisma to app usable objects and vice versa.

## Checklist


app/src/controllers/chefs.ts
```diff
@@ -52,17 +51,15 @@ const controller = {
 
   getSubmission: async (req: Request, res: Response, next: NextFunction) => {
     try {
-      const response = await chefsService.getSubmission(req.query.formId as string, req.params.submissionId);
-      res.status(200).send(response);
+      res.status(200).send(await chefsService.getSubmission(req.query.formId as string, req.params.submissionId));
```
jujaga, line 54: - `req.query.formId` should be getting pre-processed by the `addDashesToUuid` utility instead of being implicit casted into a string. Eventual Joi validation layer should be catching whether it's a valid uuidv4 at all or not down the line but for now need to make sure we're not just accepting potentially malicious strings as inputs and having it go straight into the service layer.
- I have mixed feelings about in-lining the response into the express send argument, as we generally have kept them separated out due to the need for it to async/await on a service call. While it is semantically the same, in the event there is an error, it becomes slightly harder to directly track which line triggered the fault in the stacktrace if it's all jumbled together in one line. Personally leaning towards recommending we keep it const response/send(response) style even though it's slightly more lines to keep things consistent with other controllers.

app/src/db/models/submission.ts
```diff
@@ -0,0 +1,88 @@
...
+        };
+      }
+    | {
+        disconnect: boolean;
+      };
+};
+type DBSubmission = Omit<Prisma.submissionGetPayload<typeof _submission>, 'assignedToUserId' | keyof IStamps> &
+  UserRelation;
+
+type Submission = Prisma.submissionGetPayload<typeof _submissionWithRelations>;
```
jujaga, line 28: If we ever need to start referring to these types outside of just this model DTO we'll need to move this into the types folder.

app/src/db/models/disconnectRelation.ts
jujaga, line 1: Wonder if this should just be put into a models/utils.ts instead at some point. Probably do not need to do that just yet, but if we end up needing to add another small util-like function into this directory, I'd strongly recommend grouping them together into a utils.

app/src/types/User.ts
```diff
@@ -0,0 +1,13 @@
+import { IStamps } from '../interfaces/IStamps';
+
+export type User = {
+  userId?: string; // Primary Key
```
jujaga, line 4: I guess while we're at this, we should put in the Primary Key in-line comment on all of our type models as if you were just looking at this type, it wouldn't always be immediately clear which attribute/property was behaving as a primary key.

frontend/src/components/submission/SubmissionForm.vue
```diff
@@ -53,10 +55,33 @@ const formSchema = object({
...
+    .when('intakeStatus', {
+      is: (val: string) => val !== 'SUBMITTED',
+      then: (schema) => schema.required(),
+      otherwise: (schema) => schema.notRequired()
+    })
+    .label('Assignee')
 });
 
 // Actions
+const getAssigneeOptionLabel = (e: any) => {
```
jujaga, line 69: Any type here isn't sufficient - should at bare minimum have a typing (even if in-line) that specifies that it expects this to be an object containing at least the `fullName` and `email` attributes. Ideally if we have a type for this construct already we leverage that instead.

jujaga (approved):

wilwong89 (approved):

norrisng-bc (approved):

PR #13, opened 2024-01-04
Submissions view UX/UI alignment

kyle1morel:
# Description

- Updates and aligns the submissions view with the new UX/UI design
- Adds a submissions search endpoint
- Resolves a bug in the user search endpoint

## Checklist

### DIFF
--- a/app/src/controllers/chefs.ts
+++ b/app/src/controllers/chefs.ts
@@ -6,36 +6,71 @@ import { IdentityProvider } from '../components/constants';
 
 import type { JwtPayload } from 'jsonwebtoken';
 import type { NextFunction, Request, Response } from '../interfaces/IExpress';
-import type { ChefsFormConfig, ChefsFormConfigData, ChefsSubmissionFormExport } from '../types';
+import type { ChefsFormConfig, ChefsFormConfigData, ChefsSubmissionForm, ChefsSubmissionFormExport } from '../types';
 
 const controller = {
   getFormExport: async (req: Request, res: Response, next: NextFunction) => {
     try {
       const cfg = config.get('server.chefs.forms') as ChefsFormConfig;
-      let formData = new Array<ChefsSubmissionFormExport>();
 
-      // Get a JSON export of all form data merged into a single array
-      await Promise.all(
+      const exportData: Array<Partial<ChefsSubmissionForm & { formId: string }>> = await Promise.all(
         Object.values<ChefsFormConfigData>(cfg).map(async (x: ChefsFormConfigData) => {
-          const data = await chefsService.getFormExport(x.id);
-          data.forEach((d: ChefsSubmissionFormExport) => (d.form.id = x.id));
-          formData = formData.concat(data);
+          return (await chefsService.getFormExport(x.id)).map((data: ChefsSubmissionFormExport) => {
+            const financiallySupportedValues = {
+              financiallySupportedBC: isTruthy(data.isBCHousingSupported),
+              financiallySupportedIndigenous: isTruthy(data.isIndigenousHousingProviderSupported),
+              financiallySupportedNonProfit: isTruthy(data.isNonProfitSupported),
+              financiallySupportedHousingCoop: isTruthy(data.isHousingCooperativeSupported)
+            };
+
+            return {
+              formId: x.id,
+              submissionId: data.form.submissionId,
+              confirmationId: data.form.confirmationId,
+              contactEmail: data.contactEmail,
+              contactPhoneNumber: data.contactPhoneNumber,
+              contactName: `${data.contactFirstName} ${data.contactLastName}`,
+              financiallySupported: Object.values(financiallySupportedValues).includes(true),
+              ...financiallySupportedValues,
+              intakeStatus: data.form.status,
+              latitude: data.latitude,
+              longitude: data.longitude,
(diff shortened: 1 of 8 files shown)


app/src/controllers/chefs.ts
```diff
@@ -6,44 +6,85 @@ import { IdentityProvider } from '../components/constants';
...
+              longitude: d.longitude,
+              naturalDisaster: d.naturalDisasterInd,
+              projectName: d.companyNameRegistered,
+              queuePriority: d.queuePriority,
+              singleFamilyUnits: d.singleFamilyUnits ?? d.multiFamilyUnits,
+              streetAddress: d.streetAddress,
+              submittedAt: d.form.createdAt,
+              submittedBy: d.form.username
+            });
+          });
```
jujaga, line 50: Something about this feels unsafe, especially with the confluence of Promise.all, an ignored map, and external context pushing. On a high level, it is safer to structure the logic more like the following instead (js pseudocode):
```js
const exportData = await Promise.all(Object.values.map(x => {
  return (await chefsService.getFormExport(x.id)).map(d => ({
    // The actual resolved contents you're pushing into the array
  }));
}));
```
This is safer as this guarantees order in the destination array (you are not guaranteed execution order in case there's network latency slowing something down when pushing into an array), and is generally more functional in design as you'd be bubbling up the stack and actually leveraging the await for a specific purpose.

app/src/controllers/chefs.ts
```diff
@@ -6,44 +6,85 @@ import { IdentityProvider } from '../components/constants';
...
+        submissionId: exportData.map((x) => x.submissionId as string)
+      });
+
+      // Overwrite export data with application data where possible
+      exportData = exportData.map((x: Partial<ChefsSubmissionForm & { formId: string }>) => {
+        return {
+          ...x,
+          ...result.find((y) => y?.submissionId === x.submissionId)
+        };
+      });
```
jujaga, line 65: You can abuse return closures here to avoid needing extra return statements and nesting:
```ts
      exportData = exportData.map((x: Partial<ChefsSubmissionForm & { formId: string }>) => ({
        ...x,
        ...result.find((y) => y?.submissionId === x.submissionId)
      }));
```

app/src/types/SubmissionSearchParameters.ts
```diff
@@ -0,0 +1,3 @@
+export type SubmissionSearchParameters = {
+  submissionId?: string[];
```
jujaga, line 2: Are we standardizing on Generic notation `Array<string>` or bracket notation? Need to make sure we're consistent.

app/src/controllers/chefs.ts
```diff
@@ -6,44 +6,79 @@ import { IdentityProvider } from '../components/constants';
...
-      let formData = new Array<ChefsSubmissionFormExport>();
 
-      // Get a JSON export of all form data merged into a single array
-      await Promise.all(
+      let exportData: Array<Partial<ChefsSubmissionForm & { formId: string }>> = await Promise.all(
         Object.values<ChefsFormConfigData>(cfg).map(async (x: ChefsFormConfigData) => {
-          const data = await chefsService.getFormExport(x.id);
-          data.forEach((d: ChefsSubmissionFormExport) => (d.form.id = x.id));
-          formData = formData.concat(data);
+          return (await chefsService.getFormExport(x.id)).map((data: ChefsSubmissionFormExport) => {
```
TimCsaky, line 18: i wonder if a lot of your types and functions could have simpler and clearer names and jsdocs. 
eg: the type `chefsSubmissionFormExport` could be 'chefsSubmission.
splitting the chefs calls into a separate service would also help with this.

TimCsaky (commented):
also some ui decisions i take issue with.. but overall looking good

app/src/controllers/chefs.ts
```diff
@@ -6,44 +6,79 @@ import { IdentityProvider } from '../components/constants';
...
-      );
+      ).then((x) => x.filter((y) => y.length).flat());
+
+      // Get a list of all submission IDs
+      const result = await chefsService.searchSubmissions({
+        submissionId: exportData.map((x) => x.submissionId as string)
+      });
+
+      // Overwrite export data with application data where possible
+      exportData = exportData.map((x: Partial<ChefsSubmissionForm & { formId: string }>) => ({
```
jujaga, line 56: Nitpick: May be better to assign this to a new named const variable instead of an overwrite; that way we can have exportData be a const, and subsequent use of this would be named like `mergedExportData` or something else instead. Usually better to steer away from object/primitive reassignments unless it is really needed; seems like this variable is only used one more time during the final filter and response.

jujaga (approved):

TimCsaky (approved):

app/src/controllers/chefs.ts
kyle1morel: Thanks for bringing this up. Confirms what I was thinking that some renaming is required. I've contemplated it for a bit. Will look into it in a future PR.

app/src/services/chefs.ts
```diff
@@ -105,6 +105,29 @@ const service = {
     }
   },
```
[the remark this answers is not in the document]
kyle1morel: Agreed. Will look at doing this as a separate PR - probably also including the renaming mentioned in your other comment.

PR #14, opened 2024-01-10
Front end document upload

kyle1morel:
# Description

Adds tabular view to the Submission view. New tab provides a button/drag-and-drop area for file uploading. Document uploads are done immediately, there is no confirmation button. Uploads are visible in a carousel appearing to the right of the upload area. Document upload uses COMS. A unique hash is added to the end of each filename to guarantee uniqueness.

## Types of changes

## Checklist


frontend/src/components/file/FileUpload.vue
```diff
@@ -0,0 +1,94 @@
...
+import type { Ref } from 'vue';
+
+// Props
+type Props = {
+  submissionId: string;
+};
+
+const props = withDefaults(defineProps<Props>(), {});
+
+const lastUploadedDocument = defineModel();
```
jujaga, line 19: Some kind of warning with this line:
> Prop "modelValue" should define at least its type

Maybe it's related to the defineProps thing not having a filled in type?

frontend/src/components/file/FileUpload.vue
```diff
@@ -0,0 +1,94 @@
...
+        />
+        <p class="font-bold">Upload</p>
+        <p class="mt-2 mb-0">Click or drag-and-drop</p>
+      </div>
+    </template>
+    <template #content="{ files }">
+      <div
+        v-if="files.length > 0"
+        class="flex align-items-center justify-content-center flex-column mb-3"
+      >
```
jujaga, line 66: Wondering if we still want flex/flexbox based layouts here or if we should switch over to the standard grid layout to avoid potential future layout concerns. Should be fine for PoC level however.

frontend/src/services/documentService.ts
```diff
@@ -1,5 +1,8 @@
+import { v4 } from 'uuid';
```
jujaga, line 1: Usually a safer consideration to have this be aliased to something like uuidv4 or similar so that we know at a glance what the intent is.

frontend/src/views/SubmissionView.vue
```diff
@@ -42,28 +59,88 @@ async function onSubmit(data: any) {
...
+          />
+        </div>
+        <div class="col-10 p-0">
+          <Carousel
+            v-if="documents.length"
+            :value="documents"
+            :num-visible="3"
+            :num-scroll="1"
+            :responsive-options="responsiveOptions"
+          >
```
jujaga, line 101: There's a chance we can reduce the amount of DOM element nesting by shifting the class grid stuff into the component itself with something like the following:
```vue
<FileUpload
  class="col-2"
  :submission-id="props.submissionId"
  @update:model-value="(e: Document) => documents.push(e)"
/>
<Carousel
  v-if="documents.length"
  class="col-10 p-0"
  :value="documents"
  :num-visible="3"
  :num-scroll="1"
  :responsive-options="responsiveOptions"
>
```

jujaga (approved):
Functionally works from what I can visually inspect. A couple minor optional nitpicks and lint warnings to address but looks good otherwise.

frontend/src/views/SubmissionView.vue
```diff
@@ -42,28 +59,88 @@ async function onSubmit(data: any) {
...
+    <TabPanel header="Files">
+      <div class="grid border-1 surface-border border-round text-center">
+        <div class="col-2">
+          <FileUpload
+            :submission-id="props.submissionId"
+            @update:model-value="(e: Document) => documents.push(e)"
+          />
+        </div>
+        <div class="col-10 p-0">
+          <Carousel
```
TimCsaky, line 95: youre probably going to want this upload ui in a separate component eventually.
I'm still questioning the need for a carousel at this stage and the complications that come with it, but good work figuring it out. The primevue component does look pretty good. Although I think we could adjust the size, spacing and default thumbnail lol.

frontend/src/components/file/FileUpload.vue
```diff
@@ -0,0 +1,94 @@
...
+const onFileUploadDragAndDrop = (event: FileUploadUploaderEvent) => {
+  onUpload(Array.isArray(event.files) ? event.files[0] : event.files);
+};
+
+const onFileUploadClick = () => {
+  fileInput.value.click();
+};
+
+const onUpload = async (file: File) => {
+  lastUploadedDocument.value = (
```
TimCsaky, line 37: while this is concise, if the service call fails, setting the value of lastUploadedDocument to an error or rejected promise might not be ideal. not sure.. i guess it's ok.. i'm not sure on the best pattern but i guess it's good to catch the error where it first throws

TimCsaky (approved):

frontend/src/views/SubmissionView.vue
kyle1morel: Tried. Doesn't work in this instance. Something about extraneous non props being passed that can't be inherited.

PR #21, opened 2024-01-24
Feat: Add placeholder icons for file upload preview

wilwong89:
# Description
In the screen displaying uploaded files, have a placeholder image preview of the file based on file extension.
Several SVG files uploaded
DocumentCard.vue component updated to display different svg placeholder based on extension

## Types of changes

## Checklist

- [X] I have added necessary documentation (if appropriate)


TimCsaky:
not sure why.. but an file.xlsx i uploaded is showing with the 'image' icon.

.. separate conversation but i wonder if we should base the icon on the mimeType (https://www.digipres.org/formats/mime-types/) which is available in pcns db.
i guess it's not 100 reliable but it is more accurate than the extension i believe because it looks at the encoding of the file.. where the extension can be anything.

frontend/src/utils/utils.ts
```diff
@@ -11,6 +11,51 @@ export function differential(source: any, comparer: any): any {
   return Object.fromEntries(Object.entries(source).filter(([key, value]) => comparer[key] !== value));
 }
 
+/**
+ * @function isFileCategory
+ * Compares mimeType to a list and returns a string category for that mimetype
+ * @param {string} mimeType mimeType to be categorized
+ * @returns {string} Category of file the mimeType falls under
+ */
+export function isFileCategory(mimeType: string): string {
```
kyle1morel, line 20: Would rename this to getFileCategory. is implies a comparison.

frontend/src/utils/utils.ts
```diff
@@ -11,6 +11,51 @@ export function differential(source: any, comparer: any): any {
...
+  // Checks for commonly used and likely used mimetypes
+  switch (true) {
+    case mimeType.includes('/vnd.shp'):
+    case mimeType.includes('/vnd.shx'):
+    case mimeType.includes('/dbase'):
+    case mimeType.includes('/dbf'):
+      return 'shape';
```
kyle1morel, line 27: Would make these return types constants to avoid a spelling discrepency between here and their use in DocumentCard.vue

frontend/src/utils/utils.ts
wilwong89: Renamed to getFileCategory.

frontend/src/utils/utils.ts
wilwong89: Changed it to constants as suggested, in DocumentsCard.vue as well.

kyle1morel (approved):

PR #23, opened 2024-01-31
Implement submission statistics & basic RBAC

kyle1morel:
# Description
- Create a new db function, public.get_activity_statistics()
- Adds a new tab on submissions page with a tabular data view
- Implements a basic RBAC to forbid front end navigation or api access if at least one role isn't present

## Types of changes

New feature (non-breaking change which adds functionality) -->

## Checklist


tpantella:
This is great! The table is a little wide - suggest we shrink the last two columns as wide as the column titles, first column as wide as the longest value, and squish/center it - this will buy us some blank spaces on the left and right ends outside the table to ultimately make it more readable.

if you're making it responsive we could do the above, then resizing the browser width could first keep the table the same width while reducing the blank side space, then once the table is as wide as a more narrow browser, go into the standard approach for shrinking as done elsewhere.

The last comment is numbers in tables are typically right-aligned, that might look weird until we shrink down the table but I would suggest doing that as well. Also since I couldn't test with a sample navigator, I am making the assumption that the navigator filter works.

If any of this is unclear please contact me! Just

Subin1Doo:
Hi Kyle! I think the stats page look great! It's simple and straightforward 👍 
I went through it and found some things that could be updated on top of what Tyler mentioned. 

First, the drop down calendar box isn't aligned with the text (submission by range/ month/ submissions by assigned navigator). But, I realized it gets aligned only when you select something from the drop down. I can see the boxes shifting position as I select a date. It'd be great if you could fix the position, so that it's aligned from the beginning and doesn't move. 

Another minor thing, it would be great if we could put some contrast between the statistics heading/ title vs. the actual state. Example: Submissions by intake state (as it is) : Assigned (in bold letters). This will increase the readability for the users and help them visually capture the necessary info faster :)

This is my first review comment, so please let me know if you need clarification on what I'm saying. Thank you!

frontend/src/components/submission/SubmissionStatistics.vue
```diff
@@ -0,0 +1,222 @@
...
+watch(
+  statisticFilters,
+  async () => {
+    let valid =
+      isEmpty(statisticFilters.value.userId) ||
+      (!isEmpty(statisticFilters.value.userId) &&
+        uuidValidate(statisticFilters.value.userId as string) &&
+        uuidVersion(statisticFilters.value.userId as string) === 4);
+
+    if (valid) {
```
TimCsaky, line 57: i wonder if you need this valid check.
maybe userId will either be empty or a uuid.. both work in the query.(?)

app/src/db/migrations/20231212000000_init.ts
```diff
@@ -160,6 +160,70 @@ export async function up(knex: Knex): Promise<void> {
...
+              (select count(*) from public.submission where "submittedAt" between cast(date_from as timestamp) and cast(date_to as timestamp)),
+              (select count(*) from public.submission where extract(month from cast(month_year as timestamp)) = extract(month from "submittedAt") and extract(year from cast(month_year as timestamp)) = extract(year from "submittedAt")),
+              (select count(*) from public.submission where "assignedToUserId" = user_id),
+              count(*) filter (where s."intakeStatus" = 'SUBMITTED'),
+              count(*) filter (where s."intakeStatus" = 'ASSIGNED'),
+              count(*) filter (where s."intakeStatus" = 'COMPLETE'),
+              count(*) filter (where s."applicationStatus" = 'NEW'),
+              count(*) filter (where s."applicationStatus" = 'IN PROGRESS'),
+              count(*) filter (where s."applicationStatus" = 'DELAYED'),
+              count(*) filter (where s."applicationStatus" = 'COMPLETED'),
```
wilwong89, line 210: Intake status and application status need to be title case and not upper case.

frontend/src/components/submission/SubmissionStatistics.vue
kyle1morel: userId is a uuid once a user has been selected from the drop down. Up until that point it is filled with whatever the user is typing into the box. Kind of annoying how it works.

wilwong89 (approved):

PR #25, opened 2024-02-06
feat: Add project name to page titles, display permits alphabetically

wilwong89:
-In submission page, if a project name has been added it will now display in the title, along with the confirmation ID. 
-In submissions list page, a new column for project name added 
-In perits page, permits are now sorted by the permit type's name

	modified:   frontend/src/components/submission/SubmissionForm.vue
	modified:   frontend/src/views/SubmissionView.vue
	modified:   frontend/src/views/SubmissionsView.vue

# Description

## Types of changes

## Checklist


frontend/src/views/SubmissionView.vue
```diff
@@ -55,10 +56,15 @@ async function onSubmissionSubmit(data: any) {
...
 
+const sortPermits = (): void => {
+  permits.value.sort((a: Permit, b: Permit) => a.permitType?.name.localeCompare(b.permitType?.name ?? '') || 0);
+};
+
 onMounted(async () => {
   submission.value = (await chefsService.getSubmission(props.formId, props.submissionId)).data;
   documents.value = (await documentService.listDocuments(props.submissionId)).data;
   permits.value = (await permitService.listPermits(props.submissionId)).data;
+  sortPermits();
```
kyle1morel, line 67: If we're expecting a certain initial delivery format this should be driven by the backend.

frontend/src/views/SubmissionView.vue
```diff
@@ -70,7 +76,11 @@ onMounted(async () => {
     />
     <span>Back to Submissions</span>
   </router-link>
-  <h1>Activity submission</h1>
+  <h1>
+    Activity submission:
+    <span v-if="submission?.confirmationId">{{ submission.confirmationId }}</span>
+    <span v-if="submission?.projectName">&nbsp;- {{ submission.projectName }}</span>
```
kyle1morel, line 82: CSS over HTML always. Replace &nbsp with padding

app/src/services/chefs.ts
```diff
@@ -93,6 +93,7 @@ const service = {
             submissionId: response.submission.id,
             applicationStatus: APPLICATION_STATUS_LIST.NEW,
             confirmationId: response.submission.confirmationId,
+            companyNameRegistered: response.submission.companyNameRegistered,
```
TimCsaky, line 96: i think companyNameRegistered is available at `submission.companyNameRegistered` .
I see some of the data returned from chefs is in `submission` (eg createdAt), and actual form fields nested in `submission.data`. it might be worth flattening all the attributes into one layer further up the stack to avoid confusion.

wilwong89: Good catch, thanks.

frontend/src/views/SubmissionView.vue
wilwong89: Put an orderBy in services/chefs.ts

frontend/src/views/SubmissionView.vue
wilwong89: Took out the CSS and put in a padding right in the first span. Auto formatter does something that adds an extra space if using padding left.

TimCsaky (approved):

kyle1morel (approved):

PR #26, opened 2024-02-06
Resolve issue setting updatedBy/At on update

kyle1morel:
- Adds a new global DB function **set_updatedAt()**. This simply sets the updatedAt field to the current time.
- Triggers added for each table calling this function on **before update**
- Passing correct user ID to Prisma update functions to set the updatedBy field

# Description

## Types of changes

## Checklist


app/src/db/migrations/20231212000000_init.ts
```diff
@@ -142,6 +180,16 @@ export async function up(knex: Knex): Promise<void> {
           table.unique(['permitId', 'permitTypeId', 'submissionId']);
         })
       )
+      .then(() =>
+        knex.schema.raw(`create trigger before_insert_permit_trigger
+          before insert on public.permit
```
TimCsaky, line 185: just checking, did you want to set updatedAt for inserts to this table, or is it just updates?

app/src/db/migrations/20231212000000_init.ts
```diff
@@ -510,12 +558,24 @@ export async function down(knex: Knex): Promise<void> {
...
       .then(() => knex.schema.dropTableIfExists('submission'))
+      .then(() => knex.schema.raw('DROP TRIGGER IF EXISTS before_update_user_trigger ON "user"'))
       .then(() => knex.schema.dropTableIfExists('user'))
+      .then(() =>
+        knex.schema.raw('DROP TRIGGER IF EXISTS before_update_identity_provider_trigger ON identity_provider')
+      )
       .then(() => knex.schema.dropTableIfExists('identity_provider'))
+      // Drop public schema PCNS functions
+      .then(() => knex.schema.raw('DROP FUNCTION IF EXISTS public.set_updatedAt'))
+      .then(() => knex.schema.raw('DROP FUNCTION IF EXISTS public.set_createdAt'))
```
TimCsaky, line 579: i dont see set_createdAt function anywhere.

app/src/components/utils.ts
```diff
@@ -68,6 +70,39 @@ export function getGitRevision(): string {
...
+/**
+ * @function getCurrentIdentity
+ * Attempts to acquire current identity value.
+ * Always takes first non-default value available. Yields `defaultValue` otherwise.
+ * @param {object} currentUser The express request currentUser object
+ * @param {string} [defaultValue=undefined] An optional default return value
+ * @returns {string} The current user identifier if applicable, or `defaultValue`
+ */
+export function getCurrentIdentity(currentUser: CurrentUser | undefined, defaultValue: string | undefined = undefined) {
+  return parseIdentityKeyClaims()
```
TimCsaky, line 82: i always thought this was overly complicated in COMS and i didnt like all the NIL's and defaultValue stuff.
could be one function 
```js
getCurrentUserId(currentUser){
// get claim name matching config.identityKey
// get claim value from currentUser.tokenPayload
// do database query to return userId
}
```
that being said.. if this works maybe it's better for now.

app/src/db/migrations/20231212000000_init.ts
kyle1morel: Yeah, needs to be set on insert for sorting purposes.

app/src/components/utils.ts
kyle1morel: I agree. Might try to clean it up during the refactoring process before going to prod.

app/src/db/migrations/20231212000000_init.ts
kyle1morel: Good catch. Was an old function I removed.

wilwong89 (approved):

TimCsaky (approved):

PR #28, opened 2024-02-14
feat: adding functionality for notes to be added in a log

wilwong89:
# Description
Capability to add log notes and information that records services and informations for a submission.
Backend schema, type, and API changes required for adding a new DB table.
Frontend UI, component, type, and queries related to new DB table.

## Types of changes

## Checklist

### DIFF
--- a/app/src/controllers/note.ts
+++ b/app/src/controllers/note.ts
@@ -1,0 +1,26 @@
+import { noteService } from '../services';
+
+import type { NextFunction, Request, Response } from '../interfaces/IExpress';
+import { CurrentUser } from '../types';
+
+const controller = {
+  createNote: async (req: Request, res: Response, next: NextFunction) => {
+    try {
+      const response = await noteService.createNote(req.body, req.currentUser as CurrentUser);
+      res.status(200).send(response);
+    } catch (e: unknown) {
+      next(e);
+    }
+  },
+
+  async listNotes(req: Request<{ submissionId: string }>, res: Response, next: NextFunction) {
+    try {
+      const response = await noteService.listNotes(req.params.submissionId);
+      res.status(200).send(response);
+    } catch (e: unknown) {
+      next(e);
+    }
+  }
+};
+
+export default controller;
--- a/app/src/controllers/note.ts
+++ b/app/src/controllers/note.ts
@@ -1,12 +1,19 @@
-import { noteService } from '../services';
+import { NIL } from 'uuid';
+
+import { noteService, userService } from '../services';
+import { getCurrentIdentity } from '../components/utils';
 
 import type { NextFunction, Request, Response } from '../interfaces/IExpress';
-import { CurrentUser } from '../types';
 
 const controller = {
   createNote: async (req: Request, res: Response, next: NextFunction) => {
     try {
-      const response = await noteService.createNote(req.body, req.currentUser as CurrentUser);
+      const userId = await userService.getCurrentUserId(getCurrentIdentity(req.currentUser, NIL), NIL);
+
+      // TODO: define body type in request
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const body = req.body as any;
+      const response = await noteService.createNote({ ...body, createdBy: userId });
       res.status(200).send(response);
     } catch (e: unknown) {
       next(e);
--- a/app/src/db/migrations/20231212000000_init.ts
+++ b/app/src/db/migrations/20231212000000_init.ts
@@ -205,9 +205,7 @@ export async function up(knex: Knex): Promise<void> {
           table.text('note').defaultTo('').notNullable();
           table.text('note_type').defaultTo('').notNullable();
           table.text('title').defaultTo('').notNullable();
-          table.text('createdAt');
-          table.text('createdBy');
-          table.unique(['note_id']);
+          stamps(knex, table);
         })
       )
 
--- a/app/src/db/migrations/20231212000000_init.ts
+++ b/app/src/db/migrations/20231212000000_init.ts
@@ -89,7 +89,6 @@ export async function up(knex: Knex): Promise<void> {
           table.boolean('aaiUpdated');
           table.text('waitingOn');
           table.timestamp('bringForwardDate', { useTz: true });
-          // TODO: REMOVE THIS WHEN DONE NOTE LOGS
           table.text('notes');
           table.text('intakeStatus');
           table.text('applicationStatus');
@@ -203,10 +202,11 @@ export async function up(knex: Knex): Promise<void> {
             .inTable('submission')
             .onUpdate('CASCADE')
             .onDelete('CASCADE');
-          table.text('category_type').defaultTo('').notNullable();
-          table.text('note_type').defaultTo('').notNullable();
           table.text('note').defaultTo('').notNullable();
-          stamps(knex, table);
+          table.text('note_type').defaultTo('').notNullable();
+          table.text('title').defaultTo('').notNullable();
+          table.text('createdAt');
+          table.text('createdBy');
           table.unique(['note_id']);
         })
       )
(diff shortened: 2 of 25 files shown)


app/src/controllers/note.ts
```diff
@@ -0,0 +1,26 @@
+import { noteService } from '../services';
+
+import type { NextFunction, Request, Response } from '../interfaces/IExpress';
+import { CurrentUser } from '../types';
```
kyle1morel, line 4: `import type`

app/src/services/note.ts
```diff
@@ -0,0 +1,55 @@
...
+  /**
+   * @function createNote
+   * Creates a Permit
+   * @param note Note Object
+   * @returns {Promise<object>} The result of running the findUnique operation
+   */
+  createNote: async (data: Note, currentUser: CurrentUser) => {
+    const newNote = {
+      ...data,
+      createdBy: addDashesToUuid((currentUser.tokenPayload as JwtPayload)?.idir_user_guid),
```
kyle1morel, line 19: This logic belong at the controller level

app/src/services/note.ts
```diff
@@ -0,0 +1,55 @@
...
+
+import type { CurrentUser, Note } from '../types';
+import { JwtPayload } from 'jsonwebtoken';
+
+const service = {
+  /**
+   * @function createNote
+   * Creates a Permit
+   * @param note Note Object
+   * @returns {Promise<object>} The result of running the findUnique operation
```
kyle1morel, line 14: Review all your JSDocs

app/src/db/migrations/20231212000000_init.ts
```diff
@@ -192,6 +192,31 @@ export async function up(knex: Knex): Promise<void> {
...
+            .notNullable()
+            .references('submissionId')
+            .inTable('submission')
+            .onUpdate('CASCADE')
+            .onDelete('CASCADE');
+          table.text('note').defaultTo('').notNullable();
+          table.text('note_type').defaultTo('').notNullable();
+          table.text('title').defaultTo('').notNullable();
+          table.text('createdAt');
+          table.text('createdBy');
```
kyle1morel, line 209: All tables should be using `stamps(knex, table);`

app/src/types/Note.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export type Note = {
+  noteId: string; // Primary Key
+  submissionId: string;
+  note: string;
+  noteType: string;
+  submission: ChefsSubmissionForm;
+  title: string;
+  createdAt: string | null;
+  createdBy: string | null;
```
kyle1morel, line 11: Should be merging with `Partial<IStamps>` per convention

frontend/src/components/note/NoteCard.vue
```diff
@@ -0,0 +1,101 @@
...
+        </div>
+        <div class="col-12 mb-2">
+          {{ cardData.note }}
+        </div>
+      </div>
+    </template>
+  </Card>
+</template>
+
+<style lang="scss">
```
kyle1morel, line 74: Need to use `style scoped` to prevent bleed over to other components. PrimeVue directives will need to use :deep() to properly access them once you do this.

frontend/src/components/note/NoteCard.vue
```diff
@@ -0,0 +1,101 @@
...
+// Props
+type Props = {
+  note: Note;
+  submissionId: string;
+};
+
+const props = withDefaults(defineProps<Props>(), {});
+
+// State
+const cardData: Ref<Note> = ref(props.note);
```
kyle1morel, line 20: Unnecessary state value. You are never modifying the state. Just read from the prop instead.

frontend/src/components/note/NoteModal.vue
```diff
@@ -0,0 +1,116 @@
+<script setup lang="ts">
+import { Form } from 'vee-validate';
+// import { ref } from 'vue';
+import { object, string } from 'yup';
+
+import { Dropdown, InputText, TextArea } from '@/components/form';
+import { Button, Dialog } from '@/lib/primevue';
+// import { noteService } from '@/services';
```
kyle1morel, line 8: Remove commented code

frontend/src/components/note/NoteModal.vue
```diff
@@ -0,0 +1,116 @@
...
+// resetForm is an automatic binding https://vee-validate.logaretm.com/v4/guide/components/handling-forms/
+function onSubmit(data: any, { resetForm }) {
+  if (props.note) initialFormValues = data;
+  else resetForm();
+  emit('note:submit', data);
+}
+
+onMounted(async () => {
+  // permitTypes.value = (await permitService.getPermitTypes())?.data;
+});
```
kyle1morel, line 55: Unnecessary code block

frontend/src/components/note/NoteModal.vue
```diff
@@ -0,0 +1,116 @@
...
+  title: string().required().max(100, 'Title too long').label('Title')
+});
+
+// Actions
+// @ts-expect-error TS7031
+// resetForm is an automatic binding https://vee-validate.logaretm.com/v4/guide/components/handling-forms/
+function onSubmit(data: any, { resetForm }) {
+  if (props.note) initialFormValues = data;
+  else resetForm();
+  emit('note:submit', data);
```
kyle1morel, line 50: Style: Should have a space between 49 and 50. They could be confused for the same else block

frontend/src/types/Note.ts
```diff
@@ -0,0 +1,15 @@
...
+  note: string;
+  noteType: string;
+  submissionId: string;
+  title: string;
+
+  // note_id: string; // Primary Key
+  // note: string;
+  // note_type: string;
+  // submission_id: string;
+  // title: string;
```
kyle1morel, line 14: ?

frontend/src/components/note/NoteCard.vue
wilwong89: Added `scoped` to both NoteCard and PermitCard component.

kyle1morel:
UI looks good. The notes upon save are not respecting new lines though.

kyle1morel (approved):

frontend/src/components/note/NoteCard.vue
```diff
@@ -0,0 +1,102 @@
...
+
+const props = withDefaults(defineProps<Props>(), {});
+
+// State
+const userName: Ref<string> = ref('');
+
+onMounted(async () => {
+  if (props.note.createdBy) {
+    const res = (await userService.searchUsers({ identityId: [props.note.createdBy] })).data;
+    userName.value = res.length ? res.pop().fullName : '';
```
jujaga, line 25: Is there a reason why we're using `Array.prototype.pop()` to acquire the last element's fullName (which modifies the res object by also dropping that last element) instead of a less impactful read-only `array.slice(-1)`?

frontend/src/components/note/NoteCard.vue
```diff
@@ -0,0 +1,102 @@
...
+};
+
+const props = withDefaults(defineProps<Props>(), {});
+
+// State
+const userName: Ref<string> = ref('');
+
+onMounted(async () => {
+  if (props.note.createdBy) {
+    const res = (await userService.searchUsers({ identityId: [props.note.createdBy] })).data;
```
jujaga, line 24: Optional: This could be re-written in then notation to avoid needing to make the entire onMounted callback an async function:
```ts
onMounted(() => {
  if (props.note.createdBy) {
    const res = userService.searchUsers({ identityId: [props.note.createdBy] }).then(response => response.data);
```

frontend/src/components/note/NoteCard.vue
```diff
@@ -0,0 +1,102 @@
...
+  margin: 0;
+}
+
+p {
+  margin-top: 0;
+  margin-bottom: 0;
+}
+
+.key {
+  color: #38598a;
```
jujaga, line 82: Consideration: Down the line if there is a way to reference a set of color names in the global scss that would be better than the magic constants defined here, in the event this specific color hex is used in other places.

wilwong89: Looking through the app it does seem to be used in a few places. Navbar, NoteCard, PermitCard, and SubmissionForm components uses that colour. The hexcode is from the gov-wide design reference, so it should definitely go into the global scss. But agreed that this should be done down the line.

jujaga (approved):

PR #38, opened 2024-02-29
Fix regex parsing housing unit numbers

wilwong89:
# Description
Update a regex to cover edge cases when parsing for the largest number range 

## Types of changes

## Checklist

### DIFF
--- a/app/src/controllers/submission.ts
+++ b/app/src/controllers/submission.ts
@@ -48,8 +48,18 @@ const controller = {
             const unitTypes = [data.singleFamilyUnits, data.multiFamilyUnits, data.multiFamilyUnits1];
             const maxUnits = unitTypes.reduce(
               (ac, value) => {
-                // Get max integer from value (eg: '1-49' returns 49)
-                const upperRange: number = value ? parseInt(value.toString().replace(/(.*)-/, '').trim()) : 0;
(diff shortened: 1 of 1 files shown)


app/src/controllers/submission.ts
```diff
@@ -48,8 +48,18 @@ const controller = {
...
+                // Unit types are in the form of '1-49' or '>500'
+                // .match() with regex '/(\d+)(?!.*\d)/' matches the last number in a string, puts it in array.
+                // Get max integer from last element of array.
+                let upperRange: number = 0;
+                if (value) {
+                  upperRange = parseInt(
+                    value
+                      .toString()
+                      .match(/(\d+)(?!.*\d)/)
+                      ?.pop() ?? '0'
```
jujaga, line 60: Consideration: While not major, using an array index to get the value out is slightly more performant than using the `pop()` method, especially when the output is to be a known structure/array assuming there is a match.
- `array[index]` x 33.83 ops/sec ±2.46%
- `array.pop()` x 31.80 ops/sec ±3.96%
- Ref: https://www.measurethat.net/Benchmarks/Show/23066/0/arraypop-vs-arrayindex

jujaga (approved):

PR #39, opened 2024-03-11
Additional parsing for housing unit numbers

wilwong89:
# Description

Parse housing units data such that "greaterthan[number]" value from CHEFS is replaced with ">[number]".

## Types of changes

## Checklist

### DIFF
--- a/app/src/controllers/submission.ts
+++ b/app/src/controllers/submission.ts
@@ -45,8 +45,18 @@ const controller = {
             };
 
             // Get greatest of multiple Units data
-            const unitTypes = [data.singleFamilyUnits, data.multiFamilyUnits, data.multiFamilyUnits1];
-            const maxUnits = unitTypes.reduce(
+            const unitData = [data.singleFamilyUnits, data.multiFamilyUnits, data.multiFamilyUnits1].sort();
+
+            // Replace text instances with symbols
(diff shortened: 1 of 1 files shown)


wilwong89:
On line 70, pop() vs. push(), there's achance that the regex detects multiple matching groups, with the return being an array. The last element of that array is what's needed.

kyle1morel (approved):

app/src/controllers/submission.ts
```diff
@@ -45,8 +45,18 @@ const controller = {
...
+            const unitData = [data.singleFamilyUnits, data.multiFamilyUnits, data.multiFamilyUnits1].sort();
+
+            // Replace text instances with symbols
+            const parsedUnitData = unitData.map((element) => {
+              if (typeof element === 'string') {
+                return element?.replace('greaterthan', '>');
+              } else {
+                return element;
+              }
+            });
```
jujaga, line 57: Optional: You might be able to get away with a one line ternary here:
```ts
const parsedUnitData = unitData.map((element) => typeof element === 'string' ? element?.replace('greaterthan', '>') : element);
```

jujaga (approved):

TimCsaky (approved):

PR #54, opened 2024-04-03
Include additional fields for submission table, UI updates

wilwong89:
# Description
Add the following fields to the submission table: `contactPreference`, `contactApplicantRelationship`, `projectDescription`, `isRentalUnit`.
Functionality for updating those fields.
Updated labels for certain submission types.

[SHOWCASE-3600](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3600)

## Types of changes

## Checklist


app/src/controllers/submission.ts
```diff
@@ -111,18 +111,22 @@ const controller = {
...
               ...financiallySupportedValues,
               intakeStatus: toTitleCase(data.form.status),
               locationPIDs: data.parcelID,
               latitude: data.latitude,
               longitude: data.longitude,
               naturalDisaster: data.naturalDisasterInd,
-              projectName: data.projectName,
               queuePriority: parseInt(data.queuePriority),
               singleFamilyUnits: maxUnits,
+              isRentalUnit: camelCaseToTitleCase(deDupeUnsure(data.isRentalUnit)),
```
TimCsaky, line 129: 'Unsure' is a pointless value. 
maybe the CHEFS data should be converted `true|false|null` in the database. The postgres `boolean` type .. and prisma and joi.. support that.. (eg  for knex table.boolean('isRentalUnit')
The isTruthy utils function might handle the deduping as well.

frontend/src/components/submission/SubmissionForm.vue
```diff
@@ -247,6 +261,25 @@ onBeforeMount(async () => {
...
+        name="singleFamilyUnits"
+        label="Units"
+        :disabled="!props.editable"
+      />
+      <Dropdown
+        class="col-4"
+        name="isRentalUnit"
+        label="Rental units"
+        :disabled="!props.editable"
+        :options="RentalStatusList"
```
TimCsaky, line 281: oh. i just looked at Tyler's instructions.. i guess Unsure is a value they want.
but the field should be empty by default where currently i think it will default to 'no'.. which could be a problem

TimCsaky: maybe the RentalStatusList constant just needs re-ordering so 'Unsure' is first in the array

app/src/validators/submission.ts
```diff
@@ -23,15 +23,19 @@ const schema = {
...
       contactName: Joi.string().min(0).max(255).allow(null),
+      contactApplicantRelationship: Joi.string().min(0).max(255).allow(null),
       contactPhoneNumber: Joi.string().min(0).max(255).allow(null),
       contactEmail: emailJoi,
+      contactPreference: Joi.string().min(0).max(255).allow(null),
+      projectName: Joi.string().min(0).max(255).allow(null),
+      projectDescription: Joi.string().min(0).max(255).allow(null),
       companyNameRegistered: Joi.string().min(0).max(255).allow(null),
       singleFamilyUnits: Joi.string().min(0).max(255).allow(null),
+      isRentalUnit: Joi.string().min(0).max(255).allow(null),
```
kyle1morel, line 38: This should be validated against a proper array of string values

app/src/db/migrations/20240402000000_002-submission-updates.ts
```diff
@@ -0,0 +1,23 @@
+import type { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+  return Promise.resolve().then(() =>
+    knex.schema.alterTable('submission', function (table) {
+      table.text('contact_preference');
+      table.text('contact_applicant_relationship');
+      table.text('is_rental_unit');
```
kyle1morel, line 8: Should not be nullable as we're expecting a one of a set of values.

app/src/types/Submission.ts
```diff
@@ -11,10 +11,14 @@ export type Submission = {
...
   companyNameRegistered: string | null;
   contactName: string | null;
+  contactApplicantRelationship: string | null;
   contactPhoneNumber: string | null;
   contactEmail: string | null;
+  contactPreference: string | null;
   projectName: string | null;
+  projectDescription: string | null;
   singleFamilyUnits: string | null;
+  isRentalUnit: string | null;
```
kyle1morel, line 21: Should not be nullable as we're expecting a one of a set of values.

frontend/src/components/submission/SubmissionForm.vue
kyle1morel: Would default to 'Unsure' in this case. The field also needs to be added to the form schema validation as a oneOf.

frontend/src/types/Submission.ts
```diff
@@ -9,9 +9,13 @@ export type Submission = {
   contactName?: string;
   contactPhoneNumber?: string;
   contactEmail?: string;
+  contactApplicantRelationship?: string;
   projectName?: string;
+  projectDescription?: string;
+  contactPreference?: string;
   companyNameRegistered?: string;
   singleFamilyUnits?: string;
+  isRentalUnit?: string;
```
kyle1morel, line 18: Should not be optional as we're expecting a one of a set of values.

kyle1morel (dismissed):

kyle1morel (approved):

PR #68, opened 2024-04-19
SubmissionView to store + minor general refactors

kyle1morel:
# Description

Moves the state data from the SubmissionView into a new submissionStore. All components now read and write to this store. Numerous emits have been removed as the up/down prop passing is no longer required.

PR also includes some general refactoring, and couple minor bug fixes.

a37343e Refactor NoteModal to correctly use default props
0d4658a Refactor SubmissionView + child components to use a store

## Checklist


frontend/src/components/note/NoteModal.vue
```diff
@@ -1,43 +1,46 @@
...
 
 // Default form values
 let initialFormValues: any = {
-  createdAt: new Date(),
-  bringForwardDate: null,
-  bringForwardState: null,
+  activityId: props.note?.activityId,
+  bringForwardDate: props.note?.bringForwardDate ? new Date(props.note.bringForwardDate) : null,
+  bringForwardState: props.note?.bringForwardState ?? null,
+  createdAt: props.note?.createdAt ? new Date(props.note.createdAt) : null,
```
wilwong89, line 39: If createdAt defaults to null with the date field disabled from editing on new notes, no new notes can be created as it trips up the backend validation. Validation expects a createdAt date from the frontend.

wilwong89: Using `new Date` solved the new note creation issue.

The new date is now established when the component loads and is set. So if the component was loaded at 10:45am and the navigator does a few things without navigating away and then makes a new note at 10:50am, every single note created after is still stuck at 10:45am.

frontend/src/components/note/NoteModal.vue
```diff
@@ -71,108 +74,71 @@ const formSchema = object({
...
+// resetForm is an automatic binding https://vee-validate.logaretm.com/v4/guide/components/handling-forms/
+async function onSubmit(data: any, { resetForm }) {
+  try {
+    if (!props.note) {
+      const result = (await noteService.createNote({ ...data, activityId: props.activityId })).data;
+      submissionStore.addNote(result, true);
+      resetForm();
+    } else {
+      const result = (await noteService.updateNote({ ...data, activityId: props.activityId })).data;
+      submissionStore.updateNote(props.note, result);
```
wilwong89, line 123: When a note is updated with a new date, it remains in place sorted in the frontend with the old date's value until the page is refreshed.

wilwong89 (approved):

PR #74, opened 2024-05-11
Yup validation for SHAS intake form

wilwong89:
# Description

Added frontend validation to new intake form.
Some minor UI adjustments (new error messages, margin/padding) to validation error messages

## Types of changes

## Checklist

## Further comments
PR to existing feature/intake-form branch


slhurley:
For me there is some odd validation behaviour.
- Email doesn't validate as a valid email address at all
- Lat/long didn't validate but then it did...I think it depends if you have clicked Submit or not and come back to it.
- Last verified date on the permits can't be a required field because they might not know it.
- I got a red toast after confirmation with Axios 500 so presumably the final saving after validation is not working
- The Submit button is only available on the final tab.  That "might" be ok the first time but after they submitted and failed once, if they fix something on a previous tab they are forced back to the final tab to re-submit.
- Also if the submission fails because of a required field on a previous tab, you don't get a message so you have no idea what you did wrong...it just does nothing.  You have to go back and find the red boxes.
- Is Project Name really a mandatory field?  (Subin to confirm)
- I think validation should happen on the "next tab" as well so you don't have to wait to find out you forgot something.
-

frontend/src/components/intake/ShasIntakeSchema.ts
```diff
@@ -0,0 +1,144 @@
...
+const prevYesSelected = (prevQuestion: string) => prevQuestion === BASIC_RESPONSES.YES;
+const stringRequired = string().required().max(255);
+
+export const getIntakeSchema = (ProjectLocation: Array<string>) =>
+  object({
+    applicant: object({
+      firstName: stringRequired.label('First name'),
+      lastName: stringRequired.label('Last name'),
+      phoneNumber: stringRequired.label('Phone number'),
+      email: stringRequired.label('Email'),
```
kyle1morel, line 23: Should be validating against a email regex

```ts
export const Regex = Object.freeze({
  // https://emailregex.com/
  // HTML5 - Modified to require domain of at least 2 characters
  EMAIL: '^[a-zA-Z0-9.!#$%&’*+/=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:\\.[a-zA-Z0-9-]{2,})+$'
});
```

frontend/src/components/intake/ShasIntakeSchema.ts
```diff
@@ -0,0 +1,144 @@
...
+  NumResidentialUnits,
+  ProjectRelationshipList,
+  YesNo,
+  YesNoUnsure
+} from '@/utils/constants';
+import { BASIC_RESPONSES } from '@/utils/enums';
+
+// Form validation schema
+const YesNoUnsureSchema = string().required().oneOf(YesNoUnsure);
+const prevYesSelected = (prevQuestion: string) => prevQuestion === BASIC_RESPONSES.YES;
```
kyle1morel, line 14: I'm not feeling the necessity of pulling this little bit out into it's own function?

frontend/src/components/intake/ShasIntakeSchema.ts
```diff
@@ -0,0 +1,144 @@
...
+      lastName: stringRequired.label('Last name'),
+      phoneNumber: stringRequired.label('Phone number'),
+      email: stringRequired.label('Email'),
+      relationshipToProject: string().required().oneOf(ProjectRelationshipList).label('Relationship to project'),
+      contactPreference: string().required().oneOf(ContactPreferenceList).label('Contact Preference')
+    }),
+    basic: object({
+      isDevelopedByCompanyOrOrg: string().required().oneOf(YesNo).label('Project developed'),
+      isDevelopedInBC: string().when('isDevelopedByCompanyOrOrg', {
+        is: (previousQuestion: string) => previousQuestion === BASIC_RESPONSES.YES,
```
kyle1morel, line 30: This naming threw me off for a bit. May be better to just stick to calling it `value` as there is no true context tie to the previous form value outside of the current layout. The actual value coming in is whatever is defined in the `when`.

Subin1Doo:
On top of Sharolyn's comment

- I agree with her comment on the submit button only being available on the last page, something I didn't think through.. We should make it activated on every tab
- Once they click submit and there are missing fields - can we have an alert banner? should be visible in all tabs with missing tabs (Part of my Figma design)
- once they click the page should kind of jump to the top of the page so they can see the alert banner without having to scroll
- Can the name of the tabs with missing fields change to red? (also in my Figma) If the Basic info tab had missing fields "Basic info" will turn red so they know which tabs to revisit right away
- Error messages, change font color to red and decrease font size by a little bit
- Put a bit of space between error message and type fields/ radio buttons
- Project name is a mandatory field in SHAS right now
- Probably bc this section got added today.. Natural disaster question doesn't get error message/ highlighted in red

frontend/src/components/intake/ShasIntakeSchema.ts
```diff
@@ -0,0 +1,144 @@
...
+      return (
+        context.originalValue.singleFamilySelected ||
+        context.originalValue.multiFamilySelected ||
+        context.originalValue.otherSelected
+      );
+    }),
+    location: object({
+      projectLocation: string().required().label('Location'),
+      streetAddress: string().when('projectLocation', {
+        is: (prevQuestion: string) => prevQuestion === ProjectLocation[0],
```
kyle1morel, line 99: Import the constants in this file and check directly against them. If someone were to inadvertently flip the array values around the validation would explode.

frontend/src/components/intake/ShasIntakeSchema.ts
```diff
@@ -0,0 +1,144 @@
...
+      hasAppliedProvincialPermits: string().oneOf(YesNoUnsure).required().label('Applied permits'),
+      checkProvincialPermits: string().when('hasAppliedProvincialPermits', {
+        is: (prevQuestion: string) => prevQuestion === BASIC_RESPONSES.YES || prevQuestion === BASIC_RESPONSES.UNSURE,
+        then: (schema) => schema.oneOf(YesNo).required().label('Check permits'),
+        otherwise: (schema) => schema.nullable()
+      })
+    }),
+    appliedPermits: array().of(
+      object({
+        permitTypeId: number().required().max(255).label('Permit type'),
```
kyle1morel, line 136: Should not have `max` here. While it's entirely unlikely the permit type IDs will go beyond this it could happen.

sanjaytkbabu (approved):

Subin1Doo:
Alert banner says "one of more..." It should say "one or more..."
Can we also change the font color of "...is a required field" messages to red and decrease the font size a little bit?
Has this been done? "Put a bit of space between error message and type fields/ radio buttons."

Everything else looks good to me! 

Thanks!!

frontend/src/components/intake/ShasIntakeForm.vue
```diff
@@ -161,9 +171,23 @@ onBeforeMount(async () => {
               :click-callback="clickCallback"
               title="Basic info"
               icon="fa-user"
+              :class="{
+                'p-red':
```
kyle1morel, line 175: The `p-` prefix is a primevue directive. We should not be using it in our own code. `error-red` or something would be a better fit.

frontend/src/components/intake/ShasIntakeForm.vue
```diff
@@ -302,9 +326,19 @@ onBeforeMount(async () => {
               :click-callback="clickCallback"
               title="Housing"
               icon="fa-house"
+              :class="{ 'p-red': validationErrors.includes('housing') }"
```
kyle1morel, line 329: Magic string. Should be referencing INTAKE_FORM_CATEGORIES. Same css class name comment as above.

frontend/src/components/intake/ShasIntakeForm.vue
```diff
@@ -900,9 +955,23 @@ onBeforeMount(async () => {
               :click-callback="clickCallback"
               title="Permits & Reports"
               icon="fa-file"
+              :class="{
+                'p-red':
```
kyle1morel, line 959: The `p-` prefix is a primevue directive. We should not be using it in our own code. `error-red` or something would be a better fit.

frontend/src/components/intake/ShasIntakeSchema.ts
```diff
@@ -0,0 +1,175 @@
...
+  ProjectRelationshipList,
+  Regex,
+  YesNo,
+  YesNoUnsure
+} from '@/utils/constants';
+import { BASIC_RESPONSES, INTAKE_FORM_CATEGORIES, PROJECT_LOCATION } from '@/utils/enums';
+
+// Form validation schema
+const YesNoUnsureSchema = string().required().oneOf(YesNoUnsure);
+const stringRequired = string().required().max(255);
```
kyle1morel, line 15: Naming styles should be similar. Consider `StringRequiredSchema`?

kyle1morel (approved):

PR #89, opened 2024-06-05
Permit tab - Added Last Verified Date

sanjaytkbabu:
# Description

PADS-159: Added last verified date to Permit tab
Auth status and Permit ID renamed
Set max date for Submitted date & Adjudication date
1b2e87a52a56b49e59dab41f70e203c8b283ea79 - Permit tab - added last verified date

## Types of changes

## Checklist

### DIFF
--- a/frontend/src/components/permit/PermitModal.vue
+++ b/frontend/src/components/permit/PermitModal.vue
@@ -2,7 +2,7 @@
 import { storeToRefs } from 'pinia';
 import { Form } from 'vee-validate';
 import { ref } from 'vue';
-import { object, string } from 'yup';
+import { date, object, string } from 'yup';
 
 import { Calendar, Dropdown, InputText } from '@/components/form';
 import { Button, Dialog, useConfirm, useToast } from '@/lib/primevue';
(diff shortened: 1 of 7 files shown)


frontend/src/components/permit/PermitModal.vue
```diff
@@ -221,13 +222,18 @@ async function onSubmit(data: PermitForm, { resetForm }) {
...
         <Calendar
           class="col-12 lg:col-6"
           name="adjudicationDate"
           label="Adjudication date"
         />
+        <Calendar
+          class="col-12 lg:col-6"
+          name="lastVerifiedDate"
+          label="Last verified date"
+        />
```
wilwong89, line 236: Submitting a future date fails silently on the frontend but throws error on the backend, there's no notification to the user exactly what happened. Consider adding validation that triggers an error on the frontend on future dates, or disallow future dates from being entered, or both.

slhurley:
Please add back the "last updated date" as mentioned in our discussion...that was an error in the A/C my fault.  I think it should remain the first field on the card as I believe the Navs may look for the last bit of work they did.  In fact, I believe this was the sort criteria for the cards.  (Also check with other devs on the new verified date database field that already exists in the intake form)

kyle1morel (approved):

PR #106, opened 2024-07-02
Implemented Priority Matrix for assigning Priority to Submissions

qhanson55:
# Description

Implemented Priority Matrix for assigning Priority to Submissions based on the criteria given in the [PADS-10](https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-10) ticket. Only applied functionality to new submissions and our data, not going to apply this to CHEFS data as the data isn't congruent enough to consistently assign priorities and Navigators still have the ability to manually change priorities.

## Types of changes

## Checklist


app/src/controllers/submission.ts
```diff
@@ -462,6 +469,38 @@ const controller = {
...
+      submission.multiFamilyUnits === '50-500' ||
+      submission.multiFamilyUnits === '>500' ||
+      submission.otherUnits === '50-500' ||
+      submission.otherUnits === '>500' ||
+      submission.hasRentalUnits === 'Yes' || // 2. Supports Rental Units
+      submission.financiallySupportedBC === 'Yes' || // 3. Social Housing
+      submission.financiallySupportedIndigenous === 'Yes'; // 4. Indigenous Led
+
+    const matchesPriorityTwoCriteria = // Priority 2 Criteria:
+      submission.singleFamilyUnits === '10-49' || // 1. Single Family homes w/ >9 Units
```
kyle1morel, line 492: Values should reference defined enum types.

app/tests/unit/controllers/submission.spec.ts
```diff
@@ -1244,3 +1244,138 @@ describe('updateSubmission', () => {
     expect(next).toHaveBeenCalledTimes(1);
   });
 });
+
+describe('assignPriority', () => {
```
kyle1morel, line 1248: Values should reference defined enum types.

app/src/controllers/submission.ts
qhanson55: created enums and changed to reference them.

app/tests/unit/controllers/submission.spec.ts
qhanson55: This would mean that we're just checking the enum values against the enum values. We should use the values that the submission form uses cause if they some how get changed then this should break this.

app/src/controllers/submission.ts
qhanson55: removed ones I created and added ones that we had

kyle1morel (approved):

PR #109, opened 2024-07-05
Statistics Function Updates

sanjaytkbabu:
# Description
Enquiry and financially supported stats added

## Types of changes

## Checklist

### DIFF
--- a/app/src/db/migrations/20240705184109_006_statistics-function-update.ts
+++ b/app/src/db/migrations/20240705184109_006_statistics-function-update.ts
@@ -1,0 +1,205 @@
+/* eslint-disable max-len */
+import type { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+  Promise.resolve()
+    // Drop public schema functions
+    .then(() =>
+      knex.schema.raw(`drop function if exists public.get_activity_statistics(
+      date_from text,
+      date_to text,
+      month_year text,
+      user_id uuid
+    )`)
+    )
+    // Create public schema functions
+    .then(() =>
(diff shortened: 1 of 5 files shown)


sanjaytkbabu:
> * For the financially supported, can it say "BC Housing" instead of "BC" (I see why you shortened the others but we can't on this one)
> * For the Types at the bottom, I see they are alphabetical but if it is possible it would be nice to see Guidance first, then General Enquiry then the others (kind of in the same order as they are presented to Navs in the dropdowns)

done!

Subin1Doo:
Just a minor change I'd like to ask for, for submissions at priority n, can we bold the word priority as well. So it would look like Submissions at **priority 1** and so on. Thanks!

sanjaytkbabu:
> Just a minor change I'd like to ask for, for submissions at priority n, can we bold the word priority as well. So it would look like Submissions at **priority 1** and so on. Thanks!

done!

app/src/db/migrations/20240705184109_006_statistics-function-update.ts
```diff
@@ -0,0 +1,210 @@
+import type { Knex } from 'knex';
```
kyle1morel, line 1: Please add `/* eslint-disable max-len */` as the first line in this file to prevent it from spewing warnings in this and future PRs.

Please do the same for the file 20240621180350_005_soft-delete-activity.ts

sanjaytkbabu: done

kyle1morel (approved):

PR #122, opened 2024-07-26
Implement proponent project list and project views

wilwong89:
# Description
Implement ability for proponents to see their list of submissions and the individual submission progress 

## Types of changes

## Checklist

### DIFF
--- a/app/src/utils/constants/housing.ts
+++ b/app/src/utils/constants/housing.ts
@@ -94,4 +94,10 @@ export const PROJECT_LOCATION_LIST = [ProjectLocation.LOCATION_COORDINATES, Proj
 
 export const QUEUE_PRIORITY = [1, 2, 3];
 
-export const SUBMISSION_TYPE_LIST = [SubmissionType.GUIDANCE, SubmissionType.INAPPLICABLE];
+export const SUBMISSION_TYPE_LIST = [
+  SubmissionType.ESCALATION,
+  SubmissionType.GENERAL_ENQUIRY,
+  SubmissionType.GUIDANCE,
+  SubmissionType.INAPPLICABLE,
+  SubmissionType.STATUS_REQUEST
+];
--- a/app/src/utils/constants/housing.ts
+++ b/app/src/utils/constants/housing.ts
@@ -94,10 +94,4 @@ export const PROJECT_LOCATION_LIST = [ProjectLocation.LOCATION_COORDINATES, Proj
 
 export const QUEUE_PRIORITY = [1, 2, 3];
 
-export const SUBMISSION_TYPE_LIST = [
-  SubmissionType.ESCALATION,
-  SubmissionType.GENERAL_ENQUIRY,
-  SubmissionType.GUIDANCE,
-  SubmissionType.INAPPLICABLE,
-  SubmissionType.STATUS_REQUEST
-];
+export const SUBMISSION_TYPE_LIST = [SubmissionType.GUIDANCE, SubmissionType.INAPPLICABLE];
--- a/frontend/src/components/housing/enquiry/EnquiryIntakeForm.vue
+++ b/frontend/src/components/housing/enquiry/EnquiryIntakeForm.vue
@@ -226,8 +226,6 @@ onBeforeMount(() => {
   if (props.enquiryId) loadEnquiry(props.enquiryId);
 });
 
-onMounted(() => {});
-
 async function emailConfirmation(activityId: string) {
   const configCC = getConfig.value.ches?.submission?.cc;
   const body = confirmationTemplate({
--- a/frontend/src/components/housing/enquiry/EnquiryIntakeForm.vue
+++ b/frontend/src/components/housing/enquiry/EnquiryIntakeForm.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import { storeToRefs } from 'pinia';
 import { Form } from 'vee-validate';
-import { onBeforeMount, ref, toRaw } from 'vue';
+import { onBeforeMount, onMounted, ref, toRaw } from 'vue';
 import { useRouter } from 'vue-router';
 import { object, string } from 'yup';
 
@@ -31,11 +31,13 @@ const { formUpdated, stopAutoSave } = useAutoSave(async () => {
 // Props
 type Props = {
   activityId?: string;
+  confirmationId?: string;
   enquiryId?: string;
 };
 
 const props = withDefaults(defineProps<Props>(), {
   activityId: undefined,
+  confirmationId: undefined,
   enquiryId: undefined
 });
 
@@ -223,6 +225,8 @@ async function loadEnquiry(enquiryId: string) {
 onBeforeMount(() => {
   if (props.enquiryId) loadEnquiry(props.enquiryId);
 });
+
+onMounted(() => {});
 
 async function emailConfirmation(activityId: string) {
   const configCC = getConfig.value.ches?.submission?.cc;
--- a/frontend/src/components/housing/enquiry/EnquiryIntakeForm.vue
+++ b/frontend/src/components/housing/enquiry/EnquiryIntakeForm.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import { storeToRefs } from 'pinia';
 import { Form } from 'vee-validate';
-import { onBeforeMount, onMounted, ref, toRaw } from 'vue';
+import { onBeforeMount, ref, toRaw } from 'vue';
 import { useRouter } from 'vue-router';
 import { object, string } from 'yup';
 
@@ -31,14 +31,14 @@ const { formUpdated, stopAutoSave } = useAutoSave(async () => {
 // Props
 type Props = {
   activityId?: string;
-  confirmationId?: string;
   enquiryId?: string;
+  submissionId?: string;
 };
 
 const props = withDefaults(defineProps<Props>(), {
   activityId: undefined,
-  confirmationId: undefined,
-  enquiryId: undefined
+  enquiryId: undefined,
+  submissionId: undefined
 });
 
 // Store
--- a/frontend/src/router/index.ts
+++ b/frontend/src/router/index.ts
@@ -101,7 +101,16 @@ const routes: Array<RouteRecordRaw> = [
       {
         path: 'projects',
         name: RouteName.HOUSING_PROJECTS,
-        component: () => import('@/views/housing/ProjectsView.vue'),
+        component: () => import('@/views/housing/project/ProjectListView.vue'),
+        meta: {
+          access: [Permissions.NAVIGATION_HOUSING]
+        }
+      },
+      {
+        path: 'project',
+        name: RouteName.HOUSING_PROJECT,
+        component: () => import('@/views/housing/project/ProjectView.vue'),
+        props: createProps,
         meta: {
           access: [Permissions.NAVIGATION_HOUSING]
         }
--- a/frontend/src/router/index.ts
+++ b/frontend/src/router/index.ts
@@ -99,6 +99,14 @@ const routes: Array<RouteRecordRaw> = [
         ]
       },
       {
+        path: 'projects',
+        name: RouteName.HOUSING_PROJECTS,
+        component: () => import('@/views/housing/ProjectsView.vue'),
+        meta: {
+          access: [Permissions.NAVIGATION_HOUSING]
+        }
+      },
+      {
         path: 'submissions',
         name: RouteName.HOUSING_SUBMISSIONS,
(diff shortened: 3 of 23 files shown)


naixin-zhang:
Hi Wilson, I assume the backend is completed as well for this one? I tried to add projects to the view, but couldn't complete the action. For a project to show up on Prop side, the "Assigned to" on Nav side cannot be blank. But the "Assigned to" dropdown has "no available options" in there. Could you please fix the "Assigned to" dropdown? Thanks!

app/src/utils/constants/housing.ts
```diff
@@ -94,4 +94,10 @@ export const PROJECT_LOCATION_LIST = [ProjectLocation.LOCATION_COORDINATES, Proj
...
 export const QUEUE_PRIORITY = [1, 2, 3];
 
-export const SUBMISSION_TYPE_LIST = [SubmissionType.GUIDANCE, SubmissionType.INAPPLICABLE];
+export const SUBMISSION_TYPE_LIST = [
+  SubmissionType.ESCALATION,
+  SubmissionType.GENERAL_ENQUIRY,
+  SubmissionType.GUIDANCE,
+  SubmissionType.INAPPLICABLE,
+  SubmissionType.STATUS_REQUEST
+];
```
kyle1morel, line 103: This is mixing Submission types with Enquiry types. If you need both lists you need to create a union of the two.

frontend/src/components/housing/enquiry/EnquiryIntakeForm.vue
```diff
@@ -217,6 +224,8 @@ onBeforeMount(async () => {
   };
 });
 
+onMounted(() => {});
```
kyle1morel, line 227: Empty function

frontend/src/router/index.ts
```diff
@@ -98,6 +98,23 @@ const routes: Array<RouteRecordRaw> = [
...
+      },
+      {
+        path: 'project',
+        name: RouteName.HOUSING_PROJECT,
+        component: () => import('@/views/housing/project/ProjectView.vue'),
+        props: createProps,
+        meta: {
+          access: [Permissions.NAVIGATION_HOUSING]
+        }
+      },
```
kyle1morel, line 117: Side note for us to look in to. Can vue-router do pathing like express? Can we have `project` & `project/${id}`? Then we can drop this pluralization we have to do

frontend/src/views/housing/project/ProjectView.vue
```diff
@@ -0,0 +1,384 @@
...
+    permitTypes: getPermitTypes.value
+  });
+  let secondFilter = permitFilter({
+    permitNeeded: PermitNeeded.YES,
+    permitStatus: PermitStatus.COMPLETED,
+    permits: getPermits.value,
+    permitTypes: getPermitTypes.value
+  });
+  return firstFilter.concat(secondFilter).sort(permitNameSortFcn).sort(permitBusinessSortFcn);
+});
```
kyle1morel, line 78: Possible bug here? I am seeing the same permits under Required and Submitted. 
https://pcns-dev-pr-122.apps.silver.devops.gov.bc.ca/housing/project?activityId=F734DE99&submissionId=649d5bb8-6d02-4361-bb5c-3949e9782aba

slhurley:
- The My Projects list is not populating under the correct criteria as noted on the Figma under "When to create a project in the table?".
- The layout and headers do not match the Figma at all  (different columns column names etc).  Please review the Figma and follow it exactly.

naixin-zhang:
Wilson mentioned that it might be better to hold off aligning the css with the design in Figma until after the other permit tracker ticket is worked on. However, @wilwong89 if you need to get this ticket done now, we can go through all the css change needed together to align with the design. Just pin me when you work on it.

app/src/utils/constants/housing.ts
wilwong89: Git blame shows that I somehow overwrote the correct version with the incorrect version. No idea how that happened because I don't need to do this. Reverting.

frontend/src/router/index.ts
wilwong89: Implemented as in express with path params. No more pluralization!

frontend/src/views/housing/project/ProjectView.vue
wilwong89: Something got duplicated wrong. Fixed now.

frontend/src/components/housing/enquiry/EnquiryIntakeForm.vue
wilwong89: Removed

wilwong89:
> Hi Wilson, I assume the backend is completed as well for this one? I tried to add projects to the view, but couldn't complete the action. For a project to show up on Prop side, the "Assigned to" on Nav side cannot be blank. But the "Assigned to" dropdown has "no available options" in there. Could you please fix the "Assigned to" dropdown? Thanks!
As per our Discord conversation, while it shows it is a dropdown, it actually is an autocomplete. After typing in 3 characters, it searches for a user from the user table and displays matches. Then it is possible to assign a user to submissions.

> * The My Projects list is not populating under the correct criteria as noted on the Figma under "When to create a project in the table?".
> * The layout and headers do not match the Figma at all  (different columns column names etc).  Please review the Figma and follow it exactly.
As per my notes in Discord, I misunderstood your initial question, but the project criteria is implemented correctly.  I would like to confirm that it does match the ticket AC where a project shows up only if `Submission status = "Guidance"` and `Submission AssignedTo is not empty`.
In [the ticket AC](https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-168), under page 1, point 3, it asks for a departure from headers in the ticket figma.

wilwong89:
Temporarily on hold until potential UI/UX changes are finalized.

slhurley:
Looks good.  The only thing still needing to be fixed is that the Contact info is not filled in when clicking "Ask a Navigator".

naixin-zhang:
Thanks, Wilson. The UI changes look good. Just a couple small things are missing. Could you please 
- Increase the gap below the header "Required permits" to 32px
- Increase the gap below the header "Submitted applications" to 32 px, and the gap above it to 80px.  

Aside from this, the issue of the Contact info not filled in when clicking "Ask a Navigator" is still there.

naixin-zhang:
There's something good to have, but I'm not sure if it's within the scope of this ticket, because it requires improvements on PCNS side, not PCPS. I'll still leave the comment here for you and Sharolyn to consider. 

When Navs assign themselves a project, they use the "Assigned to" field, but they rarely remember to change the Intake state to "Assigned". Can we associate the two fields, so that when "Assigned to" is filled, the Intake state is automatically changed to "Assigned"? 

There're a couple more similar cases:
- If the Authorization status is "In review" or "Pending", then the Permit state should be "Applied" automatically;
- If the Authorization status is "Issued" or "Denied", then the permit state should be "Completed" automatically.

frontend/src/views/housing/project/ProjectListView.vue
```diff
@@ -0,0 +1,95 @@
...
+import { RouteName } from '@/utils/enums/application';
+import { ApplicationStatus, IntakeStatus, SubmissionType } from '@/utils/enums/housing';
+import { submissionService } from '@/services';
+
+import type { Ref } from 'vue';
+import type { Submission } from '@/types';
+
+const router = useRouter();
+
+// Enums
```
kyle1morel, line 16: Style: These are constant lists of enums, not enums themselves

frontend/src/views/housing/project/ProjectView.vue
```diff
@@ -0,0 +1,436 @@
...
+        return 'Submitted';
+    }
+  }
+  if (permitState === PermitStatus.COMPLETED) {
+    switch (authStatus) {
+      case PermitAuthorizationStatus.ISSUED:
+      case PermitAuthorizationStatus.DENIED:
+        return authStatus;
+      default:
+        return 'Completed';
```
kyle1morel, line 125: Could use `PermitStatus.COMPLETED` here to avoid a magic string. Not much to be done about the two Submitted strings unless we want to use the `IntakeStatus` enum there?

frontend/src/views/housing/project/ProjectView.vue
```diff
@@ -0,0 +1,436 @@
...
+    >
+      {{ getSubmission?.projectName }}
+    </span>
+    <span
+      v-else
+      class="font-bold"
+    >
+      ...loading
+    </span>
+  </div>
```
kyle1morel, line 181: Is it possible to use https://primevue.org/breadcrumb/ here?

frontend/src/views/housing/project/ProjectView.vue
```diff
@@ -0,0 +1,436 @@
...
+  text-decoration: none;
+}
+
+h1 {
+  font-size: 2.125rem;
+}
+
+h3 {
+  font-size: 1.75rem;
+}
```
kyle1morel, line 312: Was it requested by the UX team to override the header sizes here?

wilwong89: Yes, they wanted to follow the BCGov design guideline stuff here. I suggested we may need to follow up for the other pages, but that it would be another time. Future pages they'll likely want to override too.

frontend/src/views/housing/project/ProjectView.vue
wilwong89: I'm not certain we should. While `IntakeStatus` does have a 'Submitted' state and string, it's technically not about `IntakeStatus` On the very rare, off chance it needs to change it would affect here.

kyle1morel (approved):

frontend/src/utils/utils.ts
```diff
@@ -244,3 +244,18 @@ export function setEmptyStringsToNull(data: any) {
...
   return data;
 }
+
+/**
+ * @function toKebabCase
+ * Converts string values to kebab case
+ * @param  {string} input The object to change
+ * @returns {string} The object with the remapped values
+ */
+export function toKebabCase(input: string | undefined | '') {
```
jujaga, line 254: Typing: The specific type `''` is already a member of type `string` and doesn't need to be there. If you need to explicitly specify empty strings, that could be something the JSDoc field could mention in lieu.

frontend/src/views/housing/project/ProjectView.vue
```diff
@@ -0,0 +1,407 @@
...
+    switch (authStatus) {
+      case PermitAuthorizationStatus.ISSUED:
+      case PermitAuthorizationStatus.DENIED:
+        return authStatus;
+      default:
+        return PermitStatus.COMPLETED;
+    }
+  }
+  return 'Submitted';
+}
```
jujaga, line 137: Optional Style: This function might be something that could be redefined as a constant attribute to value mapping set to help reduce cognitive load interpreting the conditional switch cases. It'll generally be easier to comprehend at a glance explicit enumerations of `permitState` x `authStatus` and their => result in an array (and down the line unit testable) than having it be determined by switch cases and some cascading behavior.

frontend/src/views/housing/project/ProjectView.vue
```diff
@@ -0,0 +1,407 @@
...
+  const permitsValue = (await permitService.listPermits(submissionValue.activityId)).data;
+  submissionStore.setSubmission(submissionValue);
+  submissionStore.setPermits(permitsValue);
+  typeStore.setPermitTypes(permitTypesValue);
+
+  if (submissionValue?.assignedUserId) {
+    assignee.value = (await userService.searchUsers({ userId: [submissionValue.assignedUserId] })).data[0];
+  }
+  loading.value = false;
+});
```
jujaga, line 153: Does this onMounted callback function have any form of error handling or are we expecting this to be thrown and caught up higher the vue stack? This is an async function, so if something bombs out in this body, you will get undefined behavior in the SPA application.

frontend/src/utils/utils.ts
wilwong89: Fixed.

frontend/src/views/housing/project/ProjectView.vue
wilwong89: I think my original approach did use a attribute to value mapping. However, I can no longer recall why that approach was not suitable. It was over three weeks ago and I am hesitant to change working and tested code. Perhaps for a future update.

frontend/src/views/housing/project/ProjectView.vue
wilwong89: Add a try catch handlers to redirect to a list page in case of errors in the submissionService, listPermits and permitService calls. Redirect on submission service call, leave on the project page for permit errors. Have a toast informing of issues for both.

kyle1morel (approved):

PR #125, opened 2024-08-03
Enable other identity providers to submit intakes

wilwong89:
# Description
Previously, only IDIR IDPs could submit enquiry and submissions. This has now been updated.
Requires an update to config files.

## Types of changes

## Checklist


app/src/types/IdpObjects.ts
```diff
@@ -0,0 +1,6 @@
+export type IdpObjects = {
```
kyle1morel, line 1: How do you feel about `IdpAttributes`? As these are pre filled idp information, and data coming from the SSO attributes object.

app/src/utils/utils.ts
```diff
@@ -129,6 +129,24 @@ export function getCurrentTokenClaim(
...
+export function getCurrentTokenUsername(currentUser: CurrentUser | undefined): string {
+  if (currentUser?.tokenPayload) {
+    const idpList = readIdpList();
+    const payload = currentUser.tokenPayload as JwtPayload;
+
+    const usernameKey = idpList.find((x) => x.idp === payload.identity_provider)?.username;
+
+    if (usernameKey && usernameKey in payload) return payload[usernameKey];
+  }
+  return 'UnknownUser';
```
kyle1morel, line 147: Should we return `undefined` instead? Easier to check if the parse failed and throw an error.

kyle1morel (approved):

PR #126, opened 2024-08-07
Fixing several bugs in enquiry intake and submission intake forms

wilwong89:
# Description
A variety of bug fixes for the enquiry intake and submission intake forms:
* Removing Submission Assistance where unnecessary
* Removing ability to edit certain form parts when it is read-only
* Remove unnecessary confirmation modals when navigating with the back button
* Adjusting form validation so it does not trigger prematurely

[PADS-255](https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-255)
[PADS-257](https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-257)

## Types of changes

## Checklist

### DIFF
--- a/frontend/src/components/housing/enquiry/EnquiryIntakeForm.vue
+++ b/frontend/src/components/housing/enquiry/EnquiryIntakeForm.vue
@@ -1,8 +1,8 @@
 <script setup lang="ts">
 import { storeToRefs } from 'pinia';
 import { Form } from 'vee-validate';
-import { nextTick, onBeforeMount, ref, toRaw } from 'vue';
-import { onBeforeRouteUpdate, useRouter } from 'vue-router';
+import { onBeforeMount, ref, toRaw } from 'vue';
+import { useRouter } from 'vue-router';
 import { object, string } from 'yup';
 
 import BackButton from '@/components/common/BackButton.vue';
@@ -220,19 +220,6 @@ async function loadEnquiry(enquiryId: string) {
     }
   });
 }
-
-onBeforeRouteUpdate(async (to) => {
-  if (!formRef.value) return;
-
-  if ('enquiryId' in to.query) {
-    loadEnquiry(to.query.enquiryId as string);
-  } else {
-    isFormReset.value = true;
-    await nextTick();
-    isFormReset.value = false;
-    editable.value = true;
-  }
-});
 
 onBeforeMount(async () => {
   if (props.enquiryId) loadEnquiry(props.enquiryId);
--- a/frontend/src/components/housing/enquiry/EnquiryIntakeForm.vue
+++ b/frontend/src/components/housing/enquiry/EnquiryIntakeForm.vue
@@ -46,7 +46,6 @@ const { getConfig } = storeToRefs(useConfigStore());
 const assignedActivityId: Ref<string | undefined> = ref(undefined);
 const editable: Ref<boolean> = ref(true);
 const formRef: Ref<InstanceType<typeof Form> | null> = ref(null);
-const isFormReset: Ref<boolean> = ref(false);
 const validationErrors: Ref<string[]> = ref([]);
 
 // Form validation schema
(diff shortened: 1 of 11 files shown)


slhurley:
On the Permits tab, you can still edit the existing permits if they were there.  You can change the dropdown on both the ones applied for AND the optional might need ones.  You can also still click the delete icon on the optional ones.  Otherwise I think the other bugs are fixed from my testing.

wilwong89:
> On the Permits tab, you can still edit the existing permits if they were there. You can change the dropdown on both the ones applied for AND the optional might need ones. You can also still click the delete icon on the optional ones. Otherwise I think the other bugs are fixed from my testing. Although...when going into a Draft and then immediately leaving it without changing anything it does tell you leaving will lose your unsaved changes....which might make them think their draft is not being saved. Is it possible for that message to only come up if there are unsaved changes?

Permits being editable fixed for both applied and optional ones. Also modified the back button so that if there is unsaved changes it will ask before leaving, but not ask if it has been autosaved.

frontend/src/components/housing/submission/SubmissionIntakeForm.vue
```diff
@@ -201,8 +208,18 @@ async function onSaveDraft(data: any, isAutoSave = false) {
...
   delete tempData['addressSearch'];
 
+  // Remove empty permits
+  if (Array.isArray(tempData.appliedPermits)) {
+    tempData.appliedPermits = tempData.appliedPermits.filter((x: Partial<Permit>) => x?.permitTypeId);
+  }
+  if (Array.isArray(tempData.investigatePermits)) {
+    tempData.investigatePermits = tempData.investigatePermits.filter((x: Partial<Permit>) => x?.permitTypeId);
+  }
+
```
kyle1morel, line 222: Theres a front end util function called `omit` which hides the delete ugliness. This could be rewritten as 
```ts
const tempData = omit(data, ['addressSearch']);
```

Also `tempData` should be renamed. This is the data that is actually being submitted.

frontend/src/components/housing/enquiry/EnquiryIntakeForm.vue
```diff
@@ -45,7 +46,7 @@ const { getConfig } = storeToRefs(useConfigStore());
 const assignedActivityId: Ref<string | undefined> = ref(undefined);
 const editable: Ref<boolean> = ref(true);
 const formRef: Ref<InstanceType<typeof Form> | null> = ref(null);
-const initialFormValues: Ref<any | undefined> = ref(undefined);
+const isFormReset: Ref<boolean> = ref(false);
```
kyle1morel, line 49: This is never being used. It's set as the `v-if` on the form, but the value never changes. Same goes for the enquiry form.

frontend/src/components/housing/maps/Map.vue
```diff
@@ -44,6 +46,26 @@ async function initMap() {
...
+
+function enableInteraction() {
+  map.dragging.enable();
+  map.touchZoom.enable();
+  map.doubleClickZoom.enable();
+  map.scrollWheelZoom.enable();
+  map.boxZoom.enable();
+  map.keyboard.enable();
+  if (map.tap) map.tap.enable();
+}
```
jujaga, line 67: Optional Style: With effectively two identical lists of items being permuted on, it might be worth considering jamming these map attributes into a defined array to reduce duplication and future accidental copy/pasta divergences:
```ts
const interactionElements = [map.dragging, map.touchZoom, map.doubleClickZoom, map.scrollWheelZoom, map.boxZoom, map.keyboard];

function disableInteraction() {
  interactionElements.forEach(elem => elem.disable());
  if (map.tap) map.tap.disable();
}

function enableInteraction() {
  interactionElements.forEach(elem => elem.enable());
  if (map.tap) map.tap.enable();
}
```

wilwong89: There are some Typescript issues with the map type and its constituent functions that make this cleaner approach not as useful. Will look into this at another time.

PR #134, opened 2024-08-22
Use sub from JWT

kyle1morel:
# Description

JWT now provides the correct sub. Now using that over `preffered_username`. Update code, and database to accurately reflect this change.

## Types of changes

## Checklist


app/src/services/user.ts
```diff
@@ -151,11 +151,11 @@ const service = {
    * @param {string} [defaultValue=undefined] An optional default return value
    * @returns {string} The current userId if applicable, or `defaultValue`
    */
-  getCurrentUserId: async (identityId: string, defaultValue: string | undefined = undefined) => {
+  getCurrentUserId: async (sub: string | undefined, defaultValue: string | undefined = undefined) => {
```
jujaga, line 154: Type: There should never be a situation where a subject is undefined in a JWT.
JSDoc: Don't forget to update the JSDoc param names too.

app/src/controllers/enquiry.ts
```diff
@@ -4,7 +4,7 @@ import { generateCreateStamps, generateUpdateStamps } from '../db/utils/utils';
 import { activityService, enquiryService, noteService, userService } from '../services';
 import { Initiative } from '../utils/enums/application';
 import { ApplicationStatus, IntakeStatus, NoteType, SubmissionType } from '../utils/enums/housing';
-import { getCurrentIdentity, getCurrentTokenUsername } from '../utils/utils';
+import { getCurrentTokenSub, getCurrentTokenUsername } from '../utils/utils';
```
jujaga, line 7: Optional Style: It might be possible to rename to `getCurrentSubject` and `getCurrentUsername` as this information will always be inferred to have come from the JWT token.

wilwong89 (approved):

PR #145, opened 2024-09-04
Proponent related submissions dropdown

sanjaytkbabu:
# Description
[PADS-272](https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-272)
self scope is now being checked on calling getActivityIds endpoint
hasAuthorization added to the endpoint
getActivityIds function removed from service layer and instead now using getSubmissions and filtering the data in the controller
InputText replaced with EditableDropdown in enquiry intake form for related submissions

## Types of changes

## Checklist

### DIFF
--- a/app/src/controllers/submission.ts
+++ b/app/src/controllers/submission.ts
@@ -286,9 +286,12 @@ const controller = {
 
   getActivityIds: async (req: Request<never, { self?: string }>, res: Response, next: NextFunction) => {
     try {
-      const response = await submissionService.getActivityIds();
-
-      res.status(200).json(response);
+      let response = await submissionService.getSubmissions();
+      if (req.currentAuthorization?.attributes.includes('scope:self')) {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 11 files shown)


Subin1Doo:
Project name missing, should show Activity/ confirmation ID - Project name in the drop down.
Can we also make the dropdown show when you just click on the field itself? It only shows the dropdown when you click on the arrow. Thank you!!

app/src/controllers/submission.ts
```diff
@@ -286,9 +286,12 @@ const controller = {
 
   getActivityIds: async (req: Request<never, { self?: string }>, res: Response, next: NextFunction) => {
```
kyle1morel, line 287: `self?` query param is no longer required.

kyle1morel (approved):

kyle1morel:
> Project name missing, should show Activity/ confirmation ID - Project name in the drop down. Can we also make the dropdown show when you just click on the field itself? It only shows the dropdown when you click on the arrow. Thank you!!

Added project name. As it is an editable drop down the text area has to remain clickable to enter information. The arrow is the only way to open/close it here.

PR #150, opened 2024-09-05
Proponent Confirm Email Updates

sanjaytkbabu:
# Description
[PADS-282](https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-282)
Email template updated
Enquiry Description added to the email template
html query parameter added to email endpoint
email service layer updated to pass html data to chefs depending on the content

## Types of changes

## Checklist

### DIFF
--- a/frontend/src/components/housing/enquiry/EnquiryIntakeForm.vue
+++ b/frontend/src/components/housing/enquiry/EnquiryIntakeForm.vue
@@ -25,7 +25,7 @@ import { YES_NO_LIST } from '@/utils/constants/application';
 import { CONTACT_PREFERENCE_LIST, PROJECT_RELATIONSHIP_LIST } from '@/utils/constants/housing';
 import { BasicResponse, Regex, RouteName } from '@/utils/enums/application';
 import { IntakeFormCategory, IntakeStatus } from '@/utils/enums/housing';
-import { confirmationTemplate } from '@/utils/templates';
+import { confirmationTemplateEnquiry } from '@/utils/templates';
 
 import type { Ref } from 'vue';
 import type { IInputEvent } from '@/interfaces';
@@ -200,7 +200,7 @@ async function onSubmit(data: any) {
       formRef.value?.setFieldValue('activityId', enquiryResponse.data.activityId);
       formRef.value?.setFieldValue('enquiryId', enquiryResponse.data.enquiryId);
       // Send confirmation email
-      emailConfirmation(enquiryResponse.data.activityId);
+      emailConfirmation(enquiryResponse.data.activityId, enquiryResponse.data.enquiryId);
       stopAutoSave();
     } else {
       throw new Error('Failed to retrieve correct enquiry draft data');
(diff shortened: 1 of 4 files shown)


Subin1Doo:
It's looking great! I think we should decrease the size of the logo and add the same margin to the right of the confirmation email message. Thank you!

wilwong89:
There's a issue that stems from a prior change that we should fix in this PR.
The Submission Assistance flow, because it uses `onSaveDraft` to manage the logic, does not send a confirmation email even though it in essence submits an enquiry intake and saves a submission draft.

Could you please add that logic in?

frontend/src/components/housing/enquiry/EnquiryIntakeForm.vue
```diff
@@ -256,7 +266,7 @@ async function emailConfirmation(activityId: string) {
     bodyType: 'text',
```
kyle1morel, line 266: The `bodyType` is defined here, and can be set to html as we know the email contents we wish to send. The changes to the email api don't feel necessary.

sanjaytkbabu: yes that works!

sanjaytkbabu:
> There's a issue that stems from a prior change that we should fix in this PR. The Submission Assistance flow, because it uses `onSaveDraft` to manage the logic, does not send a confirmation email even though it in essence submits an enquiry intake and saves a submission draft.
> 
> Could you please add that logic in?

Will address it in PADS-284.

Subin1Doo:
other than the assistance confirmation email mentioned above, everything is looking good!

frontend/src/utils/templates.ts
```diff
@@ -59,13 +59,64 @@ const replacePlaceholders = (baseText: string, replacementConfig: { [key: string
   return newText;
 };
 
-export const confirmationTemplate = (replaceConfig: { [key: string]: string | string[] | undefined }) => {
+export const confirmationTemplateSubmission = (replaceConfig: { [key: string]: string | string[] | undefined }) => {
+  const PCNS_URL = 'https://pcns-prod-master.apps.silver.devops.gov.bc.ca';
+  const BC_BANNER = 'https://coms.api.gov.bc.ca/api/v1/object/446ee8ee-e302-4cb8-b44d-24a1f583edba';
+  const BC_FOOTER = 'https://coms.api.gov.bc.ca/api/v1/object/853de44a-e62f-41f5-81fd-6eff6cb66d52';
```
kyle1morel, line 65: Should be a single set of constants in this file instead of duplicated in each function

frontend/src/utils/templates.ts
```diff
@@ -59,13 +59,64 @@ const replacePlaceholders = (baseText: string, replacementConfig: { [key: string
...
+    BC_FOOTER +
+    '" width="100%" alt="B.C. Government Footer" /><br><br>' +
+    '</div>';
+  return replacePlaceholders(baseTemplate, replaceConfig);
+};
+
+export const confirmationTemplateEnquiry = (replaceConfig: { [key: string]: string | string[] | undefined }) => {
+  const PCNS_URL = 'https://pcns-prod-master.apps.silver.devops.gov.bc.ca';
+  const BC_BANNER = 'https://coms.api.gov.bc.ca/api/v1/object/446ee8ee-e302-4cb8-b44d-24a1f583edba';
+  const BC_FOOTER = 'https://coms.api.gov.bc.ca/api/v1/object/853de44a-e62f-41f5-81fd-6eff6cb66d52';
```
kyle1morel, line 95: Should be a single set of constants in this file instead of duplicated in each function

frontend/src/utils/templates.ts
```diff
@@ -59,13 +59,64 @@ const replacePlaceholders = (baseText: string, replacementConfig: { [key: string
   return newText;
 };
 
-export const confirmationTemplate = (replaceConfig: { [key: string]: string | string[] | undefined }) => {
+export const confirmationTemplateSubmission = (replaceConfig: { [key: string]: string | string[] | undefined }) => {
+  const PCNS_URL = 'https://pcns-prod-master.apps.silver.devops.gov.bc.ca';
```
kyle1morel, line 63: Can you ref the url from the route instead of hard coding in the production url?

kyle1morel (approved):

PR #179, opened 2024-10-03
Improved intake form auto save handling

kyle1morel:
# Description

Moves the form autosave into a reusable component. Now watches for form value changes instead of various key inputs.

## Checklist


frontend/src/components/form/FormAutosave.vue
```diff
@@ -0,0 +1,42 @@
+<script setup lang="ts">
+import { useFormValues, useIsFormDirty } from 'vee-validate';
+import { ref, watch } from 'vue';
+
+import type { Ref } from 'vue';
+
+// Props
+const { callback, delay = 10000 } = defineProps<{
```
jujaga, line 8: Optional style: May be worth having the timeout delay be a named const value instead of as a magic number for improving readability.

jujaga (approved):

wilwong89 (commented):
Autosave toast notification triggered after navigating away from the submission intake form. We may need to use `beforeUnmount` to stop the timeout when navigating away.

kyle1morel:
> Autosave toast notification triggered after navigating away from the submission intake form. We may need to use `beforeUnmount` to stop the timeout when navigating away.

Resolved

PR #181, opened 2024-10-04
Changes to Submission/Enquiry draft saving endpoints

kyle1morel:
# Description

Submission & Enquiry draft endpoints updated to be

- `PUT /draft`
  - to keep a draft in `DRAFT` status
- `PUT /draft/submit`
  - finalize a draft and put it into `SUBMITTED` status

Param no longer required. Controller will read directly from the body and create or update data as required. 
Route query is updated on save to allow for correct browser refresh.

https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-298

## Types of changes

## Checklist

### DIFF
--- a/app/src/controllers/submission.ts
+++ b/app/src/controllers/submission.ts
@@ -497,8 +497,8 @@ const controller = {
       await permitService.deletePermitsByActivity(submission.activityId);
 
       // Create each permit
-      await Promise.all(appliedPermits.map(async (x: Permit) => await permitService.createPermit(x)));
-      await Promise.all(investigatePermits.map(async (x: Permit) => await permitService.createPermit(x)));
+      await Promise.all(appliedPermits.map((x: Permit) => permitService.createPermit(x)));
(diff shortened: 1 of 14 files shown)


app/src/controllers/submission.ts
```diff
@@ -443,18 +468,68 @@ const controller = {
...
+          ...generateCreateStamps(req.currentContext)
+        });
+      }
+
+      // Remove already existing permits for this activity
+      await permitService.deletePermitsByActivity(submission.activityId);
+
+      // Create each permit
+      await Promise.all(appliedPermits.map(async (x: Permit) => await permitService.createPermit(x)));
+      await Promise.all(investigatePermits.map(async (x: Permit) => await permitService.createPermit(x)));
```
jujaga, line 501: Concurrency: Not entirely sure if we should have the inner map callback function have an async and await in there, as that may cause issues with promise propagation back to the upper Promise.all. Might be able to have the arrow function only without the concurrent decorators.

jujaga (approved):
Semantically appears fine other than the potential async/await issue with the Promise all mapping.

PR #182, opened 2024-10-07
Barebones implementation of i18n

kyle1morel:
# Description

This is a barebones implementation of i18n to prevent a growing verbiage backlog. Existing verbiage will be moved at a future date.

## Types of changes

## Checklist


frontend/src/i18n.ts
```diff
@@ -0,0 +1,23 @@
...
+// Type-define 'en' as the master schema for the resource
+type MessageSchema = typeof en;
+
+export const SUPPORT_LOCALES = ['en'];
+
+let i18n;
+
+export function setupI18n(options = { locale: 'en' }) {
+  i18n = createI18n<[MessageSchema], 'en'>({
+    legacy: false, // Don't touch this
```
jujaga, line 14: Style: It'd be better to put "why" this shouldn't be touched like the docs
```ts
// you must set `false`, to use Composition API
```

frontend/src/locales/en.json
jujaga, line 1: Medium-term would be ideally be looking at how we can statically host these locales. Even though it seems like this is taking the javascript tree-shaking route, I have a sneaking suspicion this will still cause a bunch of potentially extra network traffic, vs managing the json resources ourselves via backend.

frontend/src/main.ts
```diff
@@ -35,6 +36,7 @@ function initializeApp(): void {
     })
   );
 
+  app.use(setupI18n());
```
jujaga, line 39: If there are no execution ordering dependencies, this might be worth introducing after Pinia and Router so that those things start loading sooner in the page load order.

frontend/src/locales/en.json
jujaga: After reviewing other larger implementations, this has been deemed an unnecessary optimization and can be shelved.

jujaga (approved):

wilwong89 (approved):

PR #193, opened 2024-11-06
Update/Clean Up User Management & Add Revoke

qhanson55:
# Description

1. Cleaned up functionality for User Management View/ Table.
2. Changed flow for Admin users so that they have one table for managing users and one for managing user access/revocation requests. See [FIGMA](https://www.figma.com/design/udtqyDmqaYte03ANaS10S2/PAT?node-id=3762-2120)
3. Added ability to revoke users
4. Cleaned up Create User Modal

These changes help with the functionality of user maangement table and solves refreshing issues, creates a cleaner flow for admins users, allows users' access to be revoked by Admins and Supervisors to request revocation, and better usability for creating users.

[PADS-257](https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-237)
[PADS-309](https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-309)

## Types of changes

## Checklist

### DIFF
--- a/app/src/routes/v1/yars.ts
+++ b/app/src/routes/v1/yars.ts
@@ -3,6 +3,7 @@ import express from 'express';
 import { yarsController } from '../../controllers';
 import { requireSomeAuth } from '../../middleware/requireSomeAuth';
 import { requireSomeGroup } from '../../middleware/requireSomeGroup';
+import { GroupName } from '../../utils/enums/application';
 
 import type { NextFunction, Request, Response } from 'express';
 
@@ -17,5 +18,12 @@ router.get('/groups', (req: Request, res: Response, next: NextFunction): void =>
 router.get('/permissions', (req: Request, res: Response, next: NextFunction): void => {
   yarsController.getPermissions(req, res, next);
 });
+
+router.delete(
+  '/subject/group',
+  (req: Request<never, never, { sub: string; group: GroupName }>, res: Response, next: NextFunction): void => {
+    yarsController.deleteSubjectGroup(req, res, next);
+  }
+);
 
 export default router;
--- a/app/src/services/accessRequest.ts
+++ b/app/src/services/accessRequest.ts
@@ -36,20 +36,6 @@ const service = {
   },
 
   /**
-   * @function deleteAccessRequests
-   * Deletes the access request
-   * @returns {Promise<object>} The result of running the delete operation
-   */
-  deleteAccessRequest: async (accessRequestId: string) => {
-    const response = await prisma.access_request.delete({
-      where: {
-        access_request_id: accessRequestId
-      }
-    });
-    return access_request.fromPrismaModel(response);
-  },
-
-  /**
    * @function getAccessRequest
    * Get an access request
    * @param {string} accessRequestId The access request data to retrieve
@@ -72,6 +58,23 @@ const service = {
   getAccessRequests: async () => {
     const response = await prisma.access_request.findMany();
     return response.map((x) => access_request.fromPrismaModel(x));
+  },
+
+  /**
+   * @function updateAccessRequest
+   * Updates a specific enquiry
+   * @param {Enquiry} data Enquiry to update
+   * @returns {Promise<Enquiry | null>} The result of running the update operation
+   */
+  updateAccessRequest: async (data: AccessRequest) => {
+    const result = await prisma.access_request.update({
+      data: { ...access_request.toPrismaModel(data), updated_at: data.updatedAt, updated_by: data.updatedBy },
+      where: {
+        access_request_id: data.accessRequestId
+      }
+    });
+
+    return access_request.fromPrismaModel(result);
   }
 };
 
--- a/frontend/src/components/user/UserCreateModal.vue
+++ b/frontend/src/components/user/UserCreateModal.vue
@@ -5,13 +5,13 @@ import { onMounted, ref } from 'vue';
 import { Dropdown } from '@/components/form';
 import { Spinner } from '@/components/layout';
 import { Button, Column, DataTable, Dialog, IconField, InputIcon, InputText, useToast } from '@/lib/primevue';
-import { ssoService } from '@/services';
+import { ssoService, yarsService } from '@/services';
 import { useAuthZStore } from '@/store';
 import { GroupName } from '@/utils/enums/application';
 
 import type { DropdownChangeEvent } from 'primevue/dropdown';
 import type { Ref } from 'vue';
-import type { User } from '@/types';
+import type { Group, User } from '@/types';
 
 // Constants
 const USER_SEARCH_PARAMS: { [key: string]: string } = {
@@ -29,9 +29,9 @@ const authzStore = useAuthZStore();
 // State
 const loading: Ref<boolean> = ref(false);
 const searchTag: Ref<string> = ref('');
-const selectableGroups: Ref<Array<GroupName>> = ref([]);
+const selectableGroups: Ref<Map<string, GroupName>> = ref(new Map());
 const selectedGroup: Ref<GroupName | undefined> = ref(undefined);
-const selectedParam: Ref<string | undefined> = ref(undefined);
+const selectedParam: Ref<string | undefined> = ref('Last name');
 const selectedUser: Ref<User | undefined> = ref(undefined);
 const users: Ref<Array<User>> = ref([]);
 const visible = defineModel<boolean>('visible');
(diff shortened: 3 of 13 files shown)


slhurley:
Is it possible to make the actual name field highlighted and ready for entry when you go into "Create new user" without having to click there first?  It also looks like you can have more than one role?  I thought they were elevated privileges and you could only have one role at a time?  I'm not sure how I did it but at one point I had a user with 3 roles showing in the list.  Also if you try to grant a role that they already have you get an Error popup but no useful error message.  Biggest issue is that it seems the User Access Requests no longer works?  The privileges between Admins and Supervisors seem to be exactly the same?

app/src/routes/v1/yars.ts
```diff
@@ -18,4 +18,8 @@ router.get('/permissions', (req: Request, res: Response, next: NextFunction): vo
   yarsController.getPermissions(req, res, next);
 });
 
+router.delete('/subgroup', (req: Request, res: Response, next: NextFunction): void => {
```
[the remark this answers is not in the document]
qhanson55: works for me

app/src/services/accessRequest.ts
```diff
@@ -72,6 +72,34 @@ const service = {
...
+  /**
+   * @function searchAccessRequests
+   * Get all access requests with given access request parameters
+   * @param {AccessRequest} accessRequest
+   * @return {Promise<Object>}
+   */
+  searchAccessRequests: async (accessRequest: AccessRequest) => {
+    const response = await prisma.access_request.findMany({ where: accessRequest });
+    return response.map((x) => access_request.fromPrismaModel(x));
+  },
```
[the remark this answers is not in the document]
qhanson55: I'm not actually using this function anymore just missed deleting it

app/src/routes/v1/yars.ts
```diff
@@ -5,6 +5,7 @@ import { requireSomeAuth } from '../../middleware/requireSomeAuth';
 import { requireSomeGroup } from '../../middleware/requireSomeGroup';
 
 import type { NextFunction, Request, Response } from 'express';
+import { GroupName } from '../../utils/enums/application';
```
kyle1morel, line 8: Import ordering

app/src/services/accessRequest.ts
```diff
@@ -72,6 +58,34 @@ const service = {
...
+  //  * @function searchAccessRequests
+  //  * Get all access requests with given access request parameters
+  //  * @param {AccessRequest} accessRequest
+  //  * @return {Promise<Object>}
+  //  */
+  // searchAccessRequests: async (accessRequest: AccessRequest) => {
+  //   const response = await prisma.access_request.findMany({ where: accessRequest });
+  //   return response.map((x) => access_request.fromPrismaModel(x));
+  // },
+
```
kyle1morel, line 73: Please remove code no longer required.

frontend/src/components/user/UserCreateModal.vue
```diff
@@ -29,12 +29,13 @@ const authzStore = useAuthZStore();
...
 const searchTag: Ref<string> = ref('');
-const selectableGroups: Ref<Array<GroupName>> = ref([]);
+const selectableGroups: Ref<Map<string, GroupName>> = ref(new Map());
 const selectedGroup: Ref<GroupName | undefined> = ref(undefined);
-const selectedParam: Ref<string | undefined> = ref(undefined);
+const selectedParam: Ref<string | undefined> = ref('Last name');
 const selectedUser: Ref<User | undefined> = ref(undefined);
 const users: Ref<Array<User>> = ref([]);
 const visible = defineModel<boolean>('visible');
+const yarsGroups: Ref<Array<Group>> = ref([]);
```
kyle1morel, line 38: Doesn't look like this is used outside of `onMounted`. I think we can removed the state value and just make it a `const` in `onMounted`.

frontend/src/components/user/UserCreateModal.vue
```diff
@@ -114,21 +135,10 @@ onMounted(() => {
             placeholder="Search by first name, last name, or email"
             class="col-12 pl-5"
             @update:model-value="searchIdirUsers"
+            autofocus
```
kyle1morel, line 138: Github warning: Attribute "autofocus" should go before "https://github.com/update:model-value"

frontend/src/components/user/UserManageModal.vue
```diff
@@ -18,19 +18,25 @@ const authzStore = useAuthZStore();
 
 // State
 const visible = defineModel<boolean>('visible');
-const selectableGroups: Ref<Array<GroupName>> = ref([]);
+const selectableGroups: Ref<Map<string, GroupName>> = ref(new Map());
 const group: Ref<GroupName | undefined> = ref(undefined);
 const yarsGroups: Ref<Array<Group>> = ref([]);
```
kyle1morel, line 23: Doesn't look like this is used outside of `onMounted`. I think we can removed the state value and just make it a `const` in `onMounted`.

frontend/src/services/yarsService.ts
```diff
@@ -11,5 +11,9 @@ export default {
 
   getPermissions(): Promise<AxiosResponse> {
     return appAxios().get(`${PATH}/permissions`);
+  },
+
+  deleteSubjectGroup(data: any): Promise<AxiosResponse> {
```
kyle1morel, line 16: We should be avoiding the use of `any` if possible.

kyle1morel:
General note, not asking for any changes at this time. New/update verbiage should be added to the `en-CA.json` file, and the text referenced via i18n. I will be asking for this to happen in future PRs.

frontend/src/components/user/UserCreateModal.vue
```diff
@@ -86,12 +86,20 @@ async function searchIdirUsers() {
   }
 }
 
-onMounted(() => {
-  // TODO: Map rbac groups to radio list to get cleaner labels
-  selectableGroups.value = [GroupName.NAVIGATOR, GroupName.NAVIGATOR_READ_ONLY];
+onMounted(async () => {
+  const yarsGroups: Array<Group> = (await yarsService.getGroups()).data;
+
+  const allowedGroups: Array<GroupName> = [GroupName.NAVIGATOR, GroupName.NAVIGATOR_READ_ONLY];
```
wilwong89, line 92: `allowedGroups` keep being defined ad-hoc in multiple files, usually with the same `[GroupName.NAVIGATOR, GroupName.NAVIGATOR_READ_ONLY]` values and modified later. Sometimes even unshifted with the same values. Could we declare this `allowedGroups` value somewhere else to be imported for consistency? Some care needs to be taken that the array imported isn't by reference when later modified with unshift.

frontend/src/components/user/UserTable.vue
```diff
@@ -2,45 +2,42 @@
...
 const selection: Ref<UserAccessRequest | undefined> = ref(undefined);
+
+function getStatusColour(data: UserAccessRequest) {
+  let colour = 'black';
+  if (data.accessRequest && data.user.status !== AccessRequestStatus.APPROVED) {
+    colour = data.accessRequest.grant ? '#036' : '#D8292F';
+  }
+  return colour;
+}
 </script>
```
wilwong89, line 33: We should not be defining hex values to be inserted into a style prop. Use Vue's ability to dynamically set CSS classes. See the <span> element below for more info.

frontend/src/components/user/UserTable.vue
```diff
@@ -69,7 +66,13 @@ const selection: Ref<UserAccessRequest | undefined> = ref(undefined);
       field="user.status"
       header="Status"
       sortable
-    />
+    >
+      <template #body="{ data }">
+        <span :style="{ color: getStatusColour(data) }">
+          {{ data.user.status }}
+        </span>
+      </template>
```
wilwong89, line 74: It's discouraged to apply styles directly to elements. Instead add a class in Vue dynamically with
`<span :class="['set-black': getStatusColour(data)]">` where `getStatusColour` would return true/false and on true would give the span the class `set-black`.

Then at the bottom of the component's scoped css, add a class with:
```
.set-black {
    color: black
}
```

Something similar can be done with the `#036` hex colour code. But instead use the css variables defined in `variables.scss` instead of hardcoding a hex value here.

frontend/src/components/user/UserTable.vue
```diff
@@ -103,8 +106,8 @@ const selection: Ref<UserAccessRequest | undefined> = ref(undefined);
         </Button>
       </template>
     </Column>
-    <Column
-      v-if="!revocation && authzStore.canNavigate(NavigationPermission.HOUSING_USER_MANAGEMENT_ADMIN)"
+    <!-- <Column
+      v-if="!requestTable && authzStore.canNavigate(NavigationPermission.HOUSING_USER_MANAGEMENT_ADMIN)"
       field="approve/deny"
```
wilwong89, line 111: Is this commented out Column required as a comment? Or can it be removed?

qhanson55: removed

frontend/src/components/user/UserTable.vue
qhanson55: changed function to return class name and added styling to file

frontend/src/components/user/UserCreateModal.vue
qhanson55: Gonna keep it as is, talked with Kyle and he agrees. Maybe could make utils function to handle this but this will be the only places this is used.

kyle1morel (approved):

PR #205, opened 2024-11-28
Intake form - advanced mapping component

sanjaytkbabu:
# Description
[PADS-160](https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-160)
Advanced mapping features added to Map component
Migration added to add geo_json column to submission table
Model and types modified to include geoJSON
SubmissionIntakeForm updated with advanced map
externalApiService updated with api calls to obtain parcel IDs on polygon draw and nearest occupant on pin drop

## Checklist


Subin1Doo:
Pin drop - We shouldn't allow them to have multiple pin drops. We should allow them to move the location of the pin drop (example: drag and drop the pin to a new location OR just clicking elsewhere will dismiss the initial pin drop and appear at the new location). 

We shouldn't allow them to pin drop and draw a polygon on the same map. It should be one or the other.  Thank you!

sanjaytkbabu:
> Pin drop - We shouldn't allow them to have multiple pin drops. We should allow them to move the location of the pin drop (example: drag and drop the pin to a new location OR just clicking elsewhere will dismiss the initial pin drop and appear at the new location).
> 
> We shouldn't allow them to pin drop and draw a polygon on the same map. It should be one or the other. Thank you!

done!

PR #211, opened 2024-12-04
Update application header

kyle1morel:
# Description

New styling. Popup menu.

https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-361

## Types of changes

## Checklist


frontend/src/components/layout/Header.vue
```diff
@@ -1,40 +1,40 @@
...
+
+// Actions
+const { t } = useI18n();
 </script>
 
 <template>
   <header>
     <nav id="header-branding">
-      <div class="flex flex-row flex-wrap align-items-center p-2 lg:pl-6">
+      <div class="flex flex-row flex-wrap align-items-center p-2 pb-0 lg:pl-6">
```
jujaga, line 13: It looks like the header height is noticeably larger than the GCPE style bars we see on the official site and Design Guide. Best I can find after some scanning is a `min-height: 65px` implicitly from their documentation example, but wasn't able to derive or find a canonical height measurement for the header element.
Ref: <https://designsystem.gov.bc.ca/react-components/?path=/docs/components-header-header--docs#controls>

I am wondering if the p-2 and pb-0 classes are causing unintended side effects on the bar height. That being said, as there doesn't seem to be a formal prescription on what a header height should precisely be, @Subin1Doo would you be able to make a determination on what the PCNS' header height should be?

jujaga (commented):
Looks nice! A few specific questions on precise representation of the header though.

frontend/src/components/layout/Header.vue
Subin1Doo: it's set to 54px in the Figma's design guide document

kyle1morel: There doesn't seem to be a set standard. https://www2.gov.bc.ca/gov/content/home uses 60px so I'm going to conform to that.

jujaga (approved):

wilwong89 (approved):

PR #213, opened 2024-12-06
Quarterly package updates

wilwong89:
# Description
Update node image to 22.12.0-alpine3.20
(Note: using alpine3.20 due to Prisma issue - https://github.com/prisma/prisma/issues/25809)
Updated several packages on both frontend and backend.
Several packages needed to be pinned to a specific version due to incompatibility issues, dependency issues. To be resolved later as these packages resolve their issues.

For both backend and frontend, removed `@tsconfig/node18` and replaced with `@tsconfig/node20`.

Backend pinned packages are:
* date-fns

Frontend pinned packages are:
* typescript
* vite-plugin-vue-devtools
* vue-tsc

https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-297

## Types of changes

## Checklist


Dockerfile
```diff
@@ -1,5 +1,5 @@
 ARG APP_ROOT=/opt/app-root/src
-ARG BASE_IMAGE=docker.io/node:20.17.0-alpine
+ARG BASE_IMAGE=docker.io/node:22.12.0-alpine
```
jujaga, line 2: > Update node image to 20.18.1-alpine

This doesn't match the PR description. Are we bumping to v22 LTS as a part of this update? If we are doing the upgrade, `@tsconfig/node22` needs to be added in lieu of the node20 variant, and `@types/node` should also correlate.

jujaga (changes requested):
Dependencies between major node versions need to be in lock step with each other. Please double check the dependency declarations and make sure we're either standardized to v20 or v22 across the board.

Dockerfile
wilwong89: Fixed PR descroption typo. updated to `@tsconfig/node22`. 
Unfortunately `@types/node` is still on v20 variant so that will need to be updated at a later time.

frontend/tsconfig.config.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": "@tsconfig/node18/tsconfig.json",
+  "extends": "@tsconfig/node20/tsconfig.json",
```
jujaga, line 2: This should be node22. Would be surprised if this didn't explode at the linting/tsc stage without node20 present in the explicit dependencies. This might be sliding by because of one of our other dependencies down the chain using node20.

wilwong89: Updated to `@tsconfig/node22` for both FE and BE.

jujaga (approved):

frontend/package.json
```diff
@@ -27,60 +27,60 @@
...
-    "vite": "^5.4.4",
-    "vite-plugin-vue-devtools": "^7.4.5",
-    "vitest": "^2.0.3",
+    "typescript": "~5.6.2",
+    "vite": "^6.0.1",
+    "vite-plugin-vue-devtools": "7.6.5",
+    "vitest": "^2.1.6",
     "volar-service-eslint": "^0.0.17",
-    "vue-tsc": "^2.1.6"
+    "vue-tsc": "~2.0.29"
```
sanjaytkbabu, line 84: looks like version has gone down, is this intentional? am I reading it wrong?

sanjaytkbabu (approved):

frontend/package.json
wilwong89: There is a temporary incompatibility between typescript and vue-tsc. There is a patch for vue-tsc that will resolve the issue next release. See this issue for details: https://github.com/vuejs/language-tools/issues/5018

qhanson55 (approved):

kyle1morel (approved):

PR #215, opened 2024-12-11
Update proponent housing view

kyle1morel:
# Description

- Updates proponent housing view to match new styling.
- Update Navbar wording and options for proponents
- Removes CPU limits from pipeline values. Recent platform change makes these no longer necessary.

https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-362
https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-400

3aa50604 Update locale text
26e10d52 Remove cpu limits
5c69a886 Update proponent housing view

## Types of changes

## Checklist


Subin1Doo:
Can we make some changes to the project list view? 
New design: Get rid of the underline for project names when in default mode. Show underline when they hover anywhere on the card. Add outline to the card when they hover anywhere on the card - outline #2E5DD7. Clicking anywhere on the card will take them to the next page. Thank you!

qhanson55 (approved):

frontend/src/assets/images/housing_2.png
jujaga, line 1: Optimization: Is there an equivalent SVG for this? For web optimization purposes if we can use SVG paths for monochrome renders like this instead of PNG that would save on network traffic.

frontend/src/views/housing/HousingView.vue
```diff
@@ -1,105 +1,251 @@
...
-          custom
+          :to="{
+            name: RouteName.HOUSING,
+            hash: '#drafts'
+          }"
+          class="no-underline"
+        >
+        {{ t('housing.drafts') }}</router-link>
+        |
+        <router-link
```
jujaga, line 92: Style: If we're attempting to create horizontal bars, this might be doable via CSS/Class `border-left` property or equivalent in lieu of using the pipe character. That might be able to make it so we don't need to use eslint and prettier ignores for this block of code.

jujaga (approved):
Seems good to me but there are few things to potentially consider.

PR #219, opened 2024-12-16
Ubuntu 24 no longer has `oc` by default

kyle1morel:
# Description

Ubuntu latest (24.04) no longer installs `oc` by default. Have to manually add this step to the pipeline. Pegging to `4.16`. This will default to the latest patch version of 4.16.

## Checklist

### DIFF
--- a/.github/workflows/on-pr-closed.yaml
+++ b/.github/workflows/on-pr-closed.yaml
@@ -30,6 +30,10 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v4
+    - name: Install oc-cli
+      uses: redhat-actions/openshift-tools-installer@v1
+      with:
+        oc: "4.16"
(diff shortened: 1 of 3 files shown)


.github/workflows/on-pr-closed.yaml
```diff
@@ -28,6 +28,10 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 12
     steps:
+    - name: Install oc
+      uses: redhat-actions/oc-installer@v1
+      with:
+        oc_version: '4.6'
```
jujaga, line 34: This should always be done after the Checkout step by convention. Do check if this can take a more recent version of OC though, 4.6 is very old already and IIRC the cluster is at 4.14 and planning on moving to 4.16 soon.

jujaga (approved):

qhanson55 (approved):

PR #230, opened 2024-12-20
Ensured and added keyboard navigation for public facing components

qhanson55:
# Description

Ensured and added the ability to navigate the application (proponent side) with keyboard
So that we are WCAG 2.2 compliant for public facing components
[PADS-310](https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-310)

## Types of changes

## Checklist

## Further comments

The tests that would cover keyboard navigation are too in depth to implement for this ticket. Should be applied to future Integration level Testing.

### DIFF
--- a/frontend/src/components/common/BackButton.vue
+++ b/frontend/src/components/common/BackButton.vue
@@ -38,14 +38,13 @@ function onConfirmLeave() {
     v-if="!confirmLeave"
     class="p-0"
     text
+    @click="router.push({ name: routeName })"
   >
-    <router-link :to="{ name: routeName }">
-      <font-awesome-icon
-        icon="fa fa-arrow-circle-left"
-        class="mr-1 app-primary-color"
-      />
-      <span class="app-primary-color">{{ text }}</span>
-    </router-link>
+    <font-awesome-icon
+      icon="fa fa-arrow-circle-left"
+      class="mr-1 app-primary-color"
+    />
+    <span class="app-primary-color">{{ text }}</span>
   </Button>
   <Button
     v-if="confirmLeave"
--- a/frontend/src/components/file/DeleteDocument.vue
+++ b/frontend/src/components/file/DeleteDocument.vue
@@ -1,4 +1,6 @@
 <script setup lang="ts">
+import { useI18n } from 'vue-i18n';
+
 import { Button, useConfirm, useToast } from '@/lib/primevue';
 import { documentService } from '@/services';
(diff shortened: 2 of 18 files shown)


slhurley:
1) First page of tracker - The project name and the icon beside it is skipped.  It should come after the breadcrumbs.  Also the tooltips are skipped.
2) Permit details page - The "What does the status mean" is skipped.

naixin-zhangbc:
**This hasn't been fixed. It still goes to the footer before the Assistance.**
> On the 2 intake forms, after the Submit button the keyboard focus should go to Assistance and then to the footer content. 

**Permit status pills' tooltips are skipped**
> All pages - Tool tip can not be accessed with keyboard

**The keyboard focus should go to the first tab-able component in that section, not the header.**
> On this page links - if you use one of the links to jump to a section, keyboard focus should go to that section. Example: Jumps to general enquiries section, pressing the tab key will take keyboard focus to submit enquiry button.

**This one is good, but I can't test it as a Proponent. So just want to make sure that the linked project name goes to the intake form on Proponent side, not on Nav side.**
> First page of tracker - The project name and the icon beside it is skipped. It should come after the breadcrumbs

frontend/src/components/common/BackButton.vue
```diff
@@ -38,14 +38,15 @@ function onConfirmLeave() {
     v-if="!confirmLeave"
     class="p-0"
     text
+    @click="router.push({ name: routeName })"
+    @keydown.enter.prevent="router.push({ name: routeName })"
```
kyle1morel, line 42: Is it possible to just do `@keydown.enter.prevent`?

frontend/src/components/file/DeleteDocument.vue
```diff
@@ -43,10 +43,12 @@ const confirmDelete = (document: Document) => {
 <template>
   <Button
     v-tooltip.bottom="'Delete document'"
+    v-tooltip.focus.bottom="'Delete document'"
```
kyle1morel, line 46: Would prefer all this text be moved to the localization file instead of duplicated.

frontend/src/components/common/BackButton.vue
kyle1morel: Misunderstood what this is doing. Ignore.

kyle1morel (approved):

kyle1morel:
Tentatively approved pending UX sign off

naixin-zhangbc:
Reviewed. All good!

PR #231, opened 2024-12-24
Update dropdown values intake forms, DB migration to convert previous values to new values

wilwong89:
# Description
Migrate several data values from old data to new data.
See tickets for details: [PADS 403](https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-403?), [PADS 412](https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-412?)
Modify CHEFS imports to account for value changes
Update dropdown values in intakes forms to account for new values.
Update tests related to intake forms.

## Types of changes

## Checklist


qhanson55:
I'm worried about using `isDevelopedByCompanyOrOrg` for Business or Individual it's not very intuitive. It may be my misunderstanding but to me `isDevelopedByBusinessOrIndividual` or something along those lines is more applicable. I think this is an old chefs value or something? If we're going to be keeping this DB as is we should try and move forward with using names that make sense to their use cases. Especially if we're doing a migration in this PR anyways.

app/src/db/migrations/20241218000000_018-intake-value-conversion.ts
jujaga, line 1: Style: Is this a migration that has to use the .raw function? On a quick glance, it doesn't seem to be using specific operations that are postgres specific. It would be better if this could still be used within the knex context if possible, even if you have to decompose the update operation into multiple subdivisions to create it.

app/src/docs/v1.api-spec.yaml
```diff
@@ -1152,7 +1152,7 @@ components:
           example: john.smith@example.com
         contactApplicantRelationship:
           type: string
-          enum: [Owner, Employee, Agent, Consultant, Other]
+          enum: [Property owner, Project consultant, Other]
```
jujaga, line 1155: Data Standards: I'm not too certain on the linkages between these actors, but I would recommend double checking to see if this enum should be changing to these descriptors with adjectives attached, as these terms may be more specific than is necessary.

app/src/docs/v1.api-spec.yaml
```diff
@@ -1640,8 +1640,9 @@ components:
           example: New
         isDevelopedByCompanyOrOrg:
           type: string
-          description: Is the project being developed by a company or organization
-          example: 'Yes'
+          enum: [Business, Individual]
+          description: Are you applying as an individual or a business?
```
jujaga, line 1644: Grammar: individual or _**as**_ a business?

app/src/docs/v1.api-spec.yaml
```diff
@@ -1994,11 +1995,10 @@ components:
...
           properties:
             isDevelopedByCompanyOrOrg:
-              type: boolean
-              description: >-
-                Indicates if the project is being developed by a company or
-                organization
-              example: true
+              type: string
+              enum: [Business, Individual]
+              description: Are you applying as an individual or a business?
```
jujaga, line 2000: Grammar: individual or _**as**_ a business?

app/src/utils/enums/housing.ts
```diff
@@ -80,11 +80,14 @@ export enum PermitStatus {
...
+  INDIVIDUAL = 'Individual'
+}
+
 export enum ProjectRelationship {
-  OWNER = 'Owner',
-  EMPLOYEE = 'Employee',
-  AGENT = 'Agent',
-  CONSULTANT = 'Consultant',
+  OWNER = 'Property owner',
+  CONSULTANT = 'Project consultant',
```
jujaga, line 90: Data Standards: I'm not too certain on the linkages between these actors, but I would recommend double checking to see if this enum should be changing to these descriptors with adjectives attached, as these terms may be more specific than is necessary.

app/src/controllers/submission.ts
```diff
@@ -174,7 +196,8 @@ const controller = {
               housingCoopDescription: data.housingCoopName,
               intakeStatus: toTitleCase(data.form.status),
               indigenousDescription: data.IndigenousHousingProviderName,
-              isDevelopedByCompanyOrOrg: toTitleCase(data.isCompany),
+              isDevelopedByCompanyOrOrg:
+                data.isCompany === 'yes' ? ProjectApplicant.BUSINESS : ProjectApplicant.INDIVIDUAL,
```
jujaga, line 200: Clarity: I would extend @qhanson55 comment w.r.t. naming clarity. This value appears to be a constrained enumeration type, and not a boolean. Usually the naming style "isSomething" implies a boolean, but this is not the case here. It might be worthwhile to consider having this attribute be more simply named as `projectApplicantType` or something for the following reasons:
- Not a boolean
- Improves semantic flexibility
  - If there's other types coming in the future like corporations or organizations, we can handle this better without the name artificially constraining it.
- Makes it more clear this is a typed value (we have an equivalent ts type definition for this already)

app/src/db/migrations/20241218000000_018-intake-value-conversion.ts
kyle1morel: I have no issues with using .raw for data updates, and has been a standard I have set in my own. It is much simpler to build the queries out in raw sql directly in DBeaver or equivalent then drop them into our migration, instead of converting them to knex. And easier for other developers to understand at a glance what is happening.

frontend/src/components/housing/submission/SubmissionIntakeForm.vue
```diff
@@ -678,9 +680,7 @@ onBeforeMount(async () => {
             </Card>
             <Card>
               <template #title>
-                <span class="section-header">
-                  Is this project being developed by a business, company, or organization?
-                </span>
+                <span class="section-header">Are you applying as an individual or a business?</span>
```
kyle1morel, line 683: Grammar: individual or _**as**_ a business? Please use localized text.

app/src/docs/v1.api-spec.yaml
kyle1morel, line 1155: Property & Project used here. Migration only has Project in the naming.

frontend/src/utils/enums/housing.ts
```diff
@@ -90,11 +90,14 @@ export enum PermitStatus {
...
+  INDIVIDUAL = 'Individual'
+}
+
 export enum ProjectRelationship {
-  OWNER = 'Owner',
-  EMPLOYEE = 'Employee',
-  AGENT = 'Agent',
-  CONSULTANT = 'Consultant',
+  OWNER = 'Property owner',
+  CONSULTANT = 'Project consultant',
```
kyle1morel, line 100: Property & Project used here. Migration only has Project in the naming.

app/src/utils/enums/housing.ts
kyle1morel, line 90: Property & Project used here. Migration only has Project in the naming.

app/src/db/migrations/20241218000000_018-intake-value-conversion.ts
wilwong89: Going to continue using .raw for now.

app/src/docs/v1.api-spec.yaml
wilwong89: After checking, PSB has given these values for us to use.

app/src/docs/v1.api-spec.yaml
wilwong89: Corrected.

app/src/docs/v1.api-spec.yaml
wilwong89: Corrected.

app/src/utils/enums/housing.ts
wilwong89: Per last comment, PSB has given these values for us to use.

app/src/controllers/submission.ts
wilwong89: After consultation, `isDevelopedByCompanyOrOrg` has been renamed to `projectApplicantType`

frontend/src/components/housing/submission/SubmissionIntakeForm.vue
wilwong89: Corrected.

app/src/docs/v1.api-spec.yaml
wilwong89: Typo corrected.

frontend/src/utils/enums/housing.ts
wilwong89: Typo corrected.

app/src/utils/enums/housing.ts
wilwong89: Typo corrected

qhanson55 (approved):

kyle1morel (approved):

PR #241, opened 2025-01-16
Added see proponents view button to nav submission view page

qhanson55:
# Description

Added see proponents view button to nav submission view page
So that navs can help proponents with the tracker if they don't understand what they're seeing

## Types of changes

## Checklist

### DIFF
--- a/frontend/src/views/housing/submission/SubmissionView.vue
+++ b/frontend/src/views/housing/submission/SubmissionView.vue
@@ -1,6 +1,7 @@
 <script setup lang="ts">
 import { storeToRefs } from 'pinia';
 import { filesize } from 'filesize';
+import { useI18n } from 'vue-i18n';
 import { computed, onMounted, ref } from 'vue';
 import { useRouter } from 'vue-router';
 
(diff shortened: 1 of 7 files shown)


sanjaytkbabu (approved):

frontend/src/views/housing/submission/SubmissionView.vue
```diff
@@ -164,23 +164,55 @@ onMounted(async () => {
...
+        })
+      "
+      @keydown.space.prevent="
+        router.push({
+          name: RouteName.HOUSING_PROJECT,
+          params: {
+            submissionId: submissionId
+          }
+        })
+      "
```
jujaga, line 210: Duplication: If we're always doing the same action across click and the accessibility options, it might be worth having a quick onViewClick arrow callback function as a singleton defined so that the attributes all call the same function instead of directly defining the router push three times.

qhanson55: addressed

kyle1morel: I'm half wondering if we want to wrap the PrimeVue button into it's own component? We leave ourselves open to missing accessibility requirements by always having to define them like this.

frontend/src/views/housing/submission/SubmissionView.vue
```diff
@@ -164,23 +173,34 @@ onMounted(async () => {
...
+      outlined
+      @click="navigateToProponentView()"
+      @keydown.enter.prevent="navigateToProponentView()"
+      @keydown.space.prevent="navigateToProponentView()"
     >
-      (Completed)
-    </span>
-  </h1>
+      <font-awesome-icon icon="fa-solid fa-eye" />
+      See Proponent's View
```
jujaga, line 201: i18n?

jujaga (approved):
Semantically makes sense. Tentatively approving with another minor comment.

wilwong89 (approved):

frontend/src/views/housing/submission/SubmissionView.vue
qhanson55: At some point Primvue has updated there button. (possibly late 3.X or in the 4 update) so this is no longer needed for PrimeVue buttons

PR #243, opened 2025-01-21
BCeID FN LN Issue Fix

sanjaytkbabu:
# Description
[PADS-436](https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-436)
BCeID was passing both fn and ln in given_name, now we're splitting it on space to fn and ln.

## Types of changes

## Checklist

### DIFF
--- a/app/src/services/user.ts
+++ b/app/src/services/user.ts
@@ -27,12 +27,15 @@ const service = {
       .filter((claims) => claims) // Drop falsy values from array
       .concat(undefined)[0]; // Set undefined as last element of array
 
+    const firstName = token.given_name.split(' ')[0];
+    const lastName = token.family_name.trim().length > 0 ? token.family_name : token.given_name.split(' ')[1];
+
     return {
       identityId: identityId,
       sub: token.sub ? token.sub : token.preferred_username,
-      firstName: token.given_name,
+      firstName: firstName,
       fullName: token.name,
-      lastName: token.family_name,
+      lastName: lastName,
       email: token.email,
       idp: token.identity_provider,
       active: true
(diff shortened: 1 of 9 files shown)


wilwong89 (approved):

qhanson55 (approved):

app/src/services/user.ts
```diff
@@ -27,12 +27,15 @@ const service = {
       .filter((claims) => claims) // Drop falsy values from array
       .concat(undefined)[0]; // Set undefined as last element of array
 
+    const firstName = token.given_name.split(' ')[0];
+    const lastName = token.family_name.trim().length > 0 ? token.family_name : token.given_name.split(' ')[1];
```
kyle1morel, line 31: If someone is named `Jim Joe Smith`, this would only store `Joe` as their last name, and `Smith` would be omitted entirely.

This will also array out of bounds in the possibility a user comes in with only a single word for their `given_name` and no `family_name`.

kyle1morel (changes requested):

jujaga (changes requested):
We cannot do this - anything we get from the JWT is canonical in nature and should not be manipulated under any circumstances when storing it in our system. We should not be interpreting the values in here because we do not have the authority to do so.

frontend/src/views/contact/ContactProfileView.vue
```diff
@@ -19,7 +20,11 @@ const formRef: Ref<InstanceType<typeof Form> | null> = ref(null);
 const initialFormValues: Ref<any | undefined> = ref(undefined);
 
 // Form validation schema
+// TODO: Sync contact key naming so common validator can be used
 const contactSchema = object({
+  firstName: string().required().max(255).label('First name'),
```
sanjaytkbabu, line 25: should we start using i18n for validation schemas?

PR #246, opened 2025-01-23
Advanced Map in Submission Intake Form

sanjaytkbabu:
# Description

## Checklist


Subin1Doo:
For the nav view of Auto Generated Location PID(s) field.. 
I noticed multiple PIDs are listed without any spacing after commas (like this ->) 000-000-000,111-111-111,222-222-222
can we put a space after commas? 
Not sure if this is possible, but can we make it so that when you double click on one of the pids it actually selects all the digits of that pid number?

sanjaytkbabu:
> For the nav view of Auto Generated Location PID(s) field.. I noticed multiple PIDs are listed without any spacing after commas (like this ->) 000-000-000,111-111-111,222-222-222 can we put a space after commas? Not sure if this is possible, but can we make it so that when you double click on one of the pids it actually selects all the digits of that pid number?

spacing added, not sure about the double click, might have to do something custom. I will check with other devs.

kyle1morel (approved):

qhanson55 (approved):

PR #249, opened 2025-01-28
Fix for critical screen reader issues, some accessibility enhancements

wilwong89:
# Description

Updated header and navbar to streamline screenreader flow
Updated submission and enquiry intake forms to have proper section headings, proper stepheader buttons
Updated permits detailed page to read out correct statuses
Fix beginning splash page header to read as one sentence
Added alt text to proponent tables

[PADS-311](https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-311?)
Remove dividers from being read out by screen reader.

## Types of changes

## Checklist


frontend/src/components/form/SectionHeader.vue
```diff
@@ -10,6 +10,6 @@ const { title } = defineProps<{
 <template>
   <div class="col-span-12 mb-4 flex items-center">
     <h4 class="flex-none m-0 pr-4">{{ title }}</h4>
-    <Divider />
+    <Divider aria-hidden="true" />
```
kyle1morel, line 13: Do we want to wrap this component into our own to ensure accessibility requirements aren't missed in the future? It appears numerous times in this PR.

kyle1morel (approved):

frontend/src/components/form/SectionHeader.vue
wilwong89: I'll make the switch. We'll still need to make sure devs remember to use the our wrapper version instead of the primevue lib.

frontend/src/views/permit/PermitStatusView.vue
```diff
@@ -355,9 +358,32 @@ onBeforeMount(async () => {
                 <div
                   class="timeline-content"
                   :class="slotProps.item.class"
+                  role="note"
+                  :aria-hidden="slotProps?.item?.iconClass == 'crossed-circle'"
                 >
                   {{ slotProps.item.text }}
                 </div>
+                <div
+                  v-if="slotProps?.item?.iconClass == 'previous'"
```
sanjaytkbabu, line 367: not sure if it's dumb, but can we have constants declared and check it here? for previous, empty and current?

sanjaytkbabu (approved):

frontend/src/views/HomeView.vue
```diff
@@ -16,8 +16,11 @@ const toHousing = (): void => {
     <div class="poly" />
     <div class="flex items-center justify-start h-full">
       <div class="text-left text-white/90">
-        <h1 class="mb-0">Welcome to the</h1>
-        <h1 class="mt-0">Permit Connect Services</h1>
+        <h1 class="mb-0">
+          Welcome to the
+          <br />
+          Permit Connect Services
+        </h1>
```
jujaga, line 23: Is this something that needs to be in i18n?

frontend/src/views/permit/PermitStatusView.vue
jujaga: The div with the v-if v-else-if stuff is making me wonder if the actual content values inside the div should be yielded as a computed or function instead and put in as a template string, instead of remounting the entire div object - it seems like all the other div attributes are basically identical otherwise.

naixin-zhangbc:
Looking great! One small change I'd like to suggest:

- When the tracker is designed to be skipped in screen reader because the application is abandoned/withdrawn/etc., can we also skip the header "Application progress"?
- I have a couple questions about other parts. Will chat about it in our 1:30 meeting.

Subin1Doo:
Things are looking great for the most part! 
Just one thing that's not working still, still can't get the screen reader to read tool-tips. 
We did mark the tooltip issue as "critical." So would be best if we can get this fixed! Thank you

frontend/src/views/permit/PermitStatusView.vue
wilwong89: Refactored to use a computed function approach.

wilwong89:
> **Per the latest discussion, we can leave these to a future ticket, and focus on fixing the tooltip issue.**
> 
> Looking great! All critical issues have been fixed. I only have small changes I'd like to suggest, but none of these should block the launch on Feb. 4.
> 
> Can we look into the followings:
> 
> * When letting the VoiceOver read out the page automatically (VO+A), it gets stuck after the header banner and the top menu bar. Users can use keyboard navigator to unstuck though.
> * On the Project page, it would be good to group the "project name, project ID, created by, navigator" information, so that screen reader reads all these information together before reading "Ask my navigator" button.
> * On permit details page, when the tracker is designed to be skipped in screen reader because the application is abandoned/withdrawn/etc., can we also skip the header "Application progress"?

> Things are looking great for the most part! Just one thing that's not working still, still can't get the screen reader to read tool-tips. We did mark the tooltip issue as "critical." So would be best if we can get this fixed! Thank you

* Unable to resolve the page readout getting stucks (VO+A), something with Vue and slots is causing the issue in the header and would require more time to find out.
* Updated the groupings to put the "Ask my navigator" button after the information.
* Added the "Application Progress" header skip on abandoned/withdrawn/etc status
* Tool tips fix is in another PR.

kyle1morel (approved):

PR #254, opened 2025-01-31
Fix tooltip stuck being displayed.

wilwong89:
# Description

Certain tooltips, when the component they are attached to are clicked, with stay stuck displayed on screen instead of disappearing after mouse leaves the attached component.
Created a custom component wrapping divs with tooltip directives to fix this issue.
[PADS-428](https://github.com/bcgov/nr-permitconnect-navigator-service/pull/254)

## Types of changes

## Checklist

### DIFF
--- a/frontend/src/components/common/Tooltip.vue
+++ b/frontend/src/components/common/Tooltip.vue
@@ -1,34 +1,74 @@
 <script setup lang="ts">
-import { useI18n } from 'vue-i18n';
-import { computed } from 'vue';
-
-// Actions
-const { t } = useI18n();
-const { down, left, right, text, up } = defineProps<{
-  down?: boolean;
+const {
+  bottom,
+  left,
+  icon = 'fa-solid fa-circle-info',
+  right,
+  text,
+  top
+} = defineProps<{
+  bottom?: boolean;
   left?: boolean;
+  icon?: string;
   right?: boolean;
   text?: string;
-  up?: boolean;
+  top?: boolean;
 }>();
-
-const direction = computed(() => {
-  if (down) return 'down';
-  else return 'up';
-});
 </script>
 
 <template>
-  <div
-    v-tooltip="{ value: text, modifier: direction }"
-    v-tooltip.focus="{ value: text, modifier: direction }"
-  >
+  <span class="tooltip-root">
+    <div class="screen-reader-only inline-block">{{ text }}</div>
+
     <div
-      v-tooltip="{ value: text, modifier: direction }"
-      v-tooltip.focus="{ value: text, modifier: direction }"
-      tabindex="0"
+      v-if="bottom"
+      v-tooltip.bottom="text"
     >
-      <font-awesome-icon icon="fa-solid fa-circle-info" />
+      <div
+        v-tooltip.focus.bottom="text"
+        tabindex="0"
+      >
+        <font-awesome-icon :icon />
+      </div>
     </div>
-  </div>
+    <div
+      v-else-if="left"
+      v-tooltip.left="text"
+    >
+      <div
+        v-tooltip.focus.left="text"
+        tabindex="0"
+      >
+        <font-awesome-icon :icon />
+      </div>
+    </div>
+    <div
+      v-else-if="right"
+      v-tooltip.right="text"
+    >
+      <div
+        v-tooltip.focus.right="text"
+        tabindex="0"
+      >
+        <font-awesome-icon :icon />
+      </div>
+    </div>
+    <div
+      v-else-if="top"
+      v-tooltip.top="text"
+    >
+      <div
+        v-tooltip.focus.top="text"
+        tabindex="0"
+      >
+        <font-awesome-icon :icon />
+      </div>
+    </div>
+  </span>
 </template>
+
+<style lang="scss" scoped>
+.tooltip-root {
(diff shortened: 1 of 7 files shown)


frontend/src/components/common/Tooltip.vue
```diff
@@ -0,0 +1,74 @@
...
+      </div>
+    </div>
+  </span>
+</template>
+
+<style lang="scss" scoped>
+.tooltip-root {
+  color: var(--p-primary-color);
+}
+</style>
```
kyle1morel, line 74: Avoid scoped css where possible. If possible, attempt to modify the custom theme. If the styling cannot be accomplished via the custom theme then consider `main.scss`. As this is simply a colour override, we already have existing classes for this in our main file.

frontend/src/components/common/Tooltip.vue
```diff
@@ -0,0 +1,74 @@
...
+      v-tooltip.top="text"
+    >
+      <div
+        v-tooltip.focus.top="text"
+        tabindex="0"
+      >
+        <font-awesome-icon :icon />
+      </div>
+    </div>
+  </span>
```
kyle1morel, line 67: Component will not render anything if one of the directional options is not provided.

frontend/src/components/housing/enquiry/EnquiryIntakeForm.vue
```diff
@@ -289,14 +290,14 @@ onBeforeMount(async () => {
...
-            v-tooltip.right="t('enquiryIntakeForm.contactTooltip')"
-            v-tooltip.focus.right="t('enquiryIntakeForm.contactTooltip')"
-            tabindex="0"
-          >
-            <font-awesome-icon icon="fa-solid fa-circle-info" />
-          </span>
+          <div class="flex">
+            <span class="section-header">{{ t('enquiryIntakeForm.contactInformation') }}</span>
+            <Tooltip
+              class="text-base flex align-items"
```
kyle1morel, line 296: This goes for all instances: Style classes like `flex` and `align` shouldn't be required for this component. The component shouldn't handle the appropriate layout internally. Style classes like `margin` or font size adjustments are what would be expected.

sanjaytkbabu:
the tooltip text is being shown outside of it.

frontend/src/components/common/Tooltip.vue
wilwong89: Updated to use the exiting colour override class.

frontend/src/components/common/Tooltip.vue
wilwong89: Default added at the template level, will default to right hand side.

frontend/src/components/housing/enquiry/EnquiryIntakeForm.vue
wilwong89: Made a change to the component and it should no longer require the style classes like `flex` and `align`. Those classes have been removed.

sanjaytkbabu (approved):

qhanson55 (commented):
The tab navigation no longer works for the tooltips it seems, tabbing over tooltip should render tooltip

wilwong89:
> The tab navigation no longer works for the tooltips it seems, tabbing over tooltip should render tooltip

I was told to not have enable tabbing over for revealing tooltip. However, it will be read by screen reader software when navigated over.

frontend/src/components/common/Tooltip.vue
```diff
@@ -0,0 +1,52 @@
+<script setup lang="ts">
+// Defaults tooltip to right if not stated.
+const {
+  bottom,
+  left,
+  icon = 'fa-solid fa-circle-info',
+  /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
```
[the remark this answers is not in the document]
wilwong89: Updated with `v-else-if="right"`

frontend/src/views/housing/HousingView.vue
```diff
@@ -118,24 +119,20 @@ onMounted(async () => {
...
-        tabindex="-1"
-      >
-        {{ t('housing.myProjects') }}
-        <div
-          v-tooltip.right="t('housing.projectsTooltip')"
-          class="flex items-center pl-2 text-xl"
+      <div class="flex items-center flex-row">
+        <h2
+          id="projects"
+          class="flex"
```
kyle1morel, line 125: Don't believe this `flex` is doing anything as their are no child elements or alignment happening.

frontend/src/views/housing/HousingView.vue
```diff
@@ -219,22 +216,21 @@ onMounted(async () => {
...
-        {{ t('housing.generalEnquiries') }}
-        <div
-          v-tooltip.right="t('housing.enquiriesTooltip')"
-          v-tooltip.focus.right="t('housing.enquiriesTooltip')"
-          class="flex items-center pl-2 text-xl"
-          tabindex="0"
+      <div class="flex items-center flex-row">
+        <h2
+          id="enquiries"
+          class="flex font-bold"
```
kyle1morel, line 222: Don't believe this `flex` is doing anything as their are no child elements or alignment happening.

wilwong89: Class removed.

qhanson55 (approved):

PR #256, opened 2025-02-04
Enable navigator enquiry form to search and add ATS clients

wilwong89:
# Description

Created database model for new enquiry fields
Updated BE and FE types
Updated navigator enquiry form UI for new feature
[PADS-366](https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-366?)

## Types of changes

## Checklist

### DIFF
--- a/app/src/db/migrations/20250203000000_022-enquiry-ats-integration.ts
+++ b/app/src/db/migrations/20250203000000_022-enquiry-ats-integration.ts
@@ -1,0 +1,20 @@
+/* eslint-disable max-len */
+import type { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+  return Promise.resolve().then(() =>
+    knex.schema.alterTable('enquiry', (table) => {
+      table.text('ats_client_number');
+      table.boolean('added_to_ats').notNullable().defaultTo(false);
+    })
+  );
+}
+
+export async function down(knex: Knex): Promise<void> {
+  return Promise.resolve().then(() =>
+    knex.schema.alterTable('enquiry', (table) => {
+      table.dropColumn('added_to_ats');
+      table.dropColumn('ats_client_number');
+    })
+  );
+}
--- a/app/src/validators/submission.ts
+++ b/app/src/validators/submission.ts
@@ -1,6 +1,7 @@
 import Joi from 'joi';
 
 import { appliedPermit } from './appliedPermit';
+import atsValidator from './ats';
 import { basicIntake } from './basic';
 import { activityId, email, uuidv4 } from './common';
 import { contacts } from './contact';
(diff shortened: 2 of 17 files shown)


app/src/validators/submission.ts
```diff
@@ -169,8 +170,9 @@ const schema = {
         .valid(...YES_NO_LIST)
         .required(),
       projectLocationDescription: Joi.string().allow(null).max(4000),
-      addedToATS: Joi.boolean().required(),
-      atsClientNumber: Joi.string().allow(null).max(255),
+      // addedToATS: Joi.boolean().required(),
```
qhanson55, line 173: need commented lines?

wilwong89: Nope! Removed.

qhanson55 (approved):

app/src/db/migrations/20250203000000_022-enquiry-ats-integration.ts
```diff
@@ -0,0 +1,20 @@
+/* eslint-disable max-len */
+import type { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+  return Promise.resolve().then(() =>
+    knex.schema.alterTable('enquiry', (table) => {
+      table.text('ats_client_number');
```
kyle1morel, line 7: ATS API returns this as a number. We should be matching the data type given to us.

frontend/src/components/user/ATSUserCreateModal.vue
```diff
@@ -40,18 +41,22 @@ const toast = useToast();
...
   try {
     loading.value = true;
+
+    const address: Partial<AddressResource> = {
+      '@type': 'AddressResource',
+      primaryPhone: submissionOrEnquiry.contacts[0]?.phoneNumber ?? '',
+      email: submissionOrEnquiry.contacts[0]?.email ?? ''
+    };
+
+    if ('streetAddress' in submissionOrEnquiry) address.addressLine1 = submissionOrEnquiry.streetAddress;
```
kyle1morel, line 51: Style: This is a little odd. Can you not just do `address.addressLine1 = submissionOrEnquiry.streetAddress`. This should just return undefined if the value doesn't exist. Or is there a validation issue on the ATS side with undefined values?

frontend/src/components/user/ATSUserCreateModal.vue
```diff
@@ -75,16 +80,20 @@ async function createATSClient() {
 }
 
 onMounted(() => {
-  const locationAddressStr = [submission.streetAddress, submission.locality, submission.province]
+  const locationAddressStr = [
```
kyle1morel, line 83: Style: Can you not use null coalescing operators here?

wilwong89:
All commits in this PR are included in the [PR-256](https://github.com/bcgov/nr-permitconnect-navigator-service/pull/256) 
Comments here will be addressed in that PR as well.

frontend/src/components/user/ATSUserCreateModal.vue
wilwong89: Typescript states an error trying to do `submissionOrEnquiry.streetAddress` as `submissionOrEnquiry` has a type that is a union between `Submission` and `Enquiry` types, but do not overlap fields. `streetAddress` is a field not shared between the two objects and the if check is for the key is the only way I found to bypass this issue.

Also ATS does not like falsey values in fields, null/undefined/empty string triggers errors in the call.

frontend/src/components/user/ATSUserCreateModal.vue
wilwong89: Same issue with typescript errors from the comment above.

app/src/db/migrations/20250203000000_022-enquiry-ats-integration.ts
wilwong89: `ats_client_number` is a text field in the `submission` table, the `enquiry` table follows suit for consistency there.

PR #262, opened 2025-02-12
added button and endpoint for downloading submission permit data

qhanson55:
# Description

Added a button in the frontend, new "reporting" endpoint, and a new rbac resource that allow the downloading of submission and permit data
Users with navigator read only, navigator, supervisor, and admin can perform this action. 
[PADS-437](https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-437)

## Types of changes

## Checklist


qhanson55:
> Can we preserve the column headers? Without this I do not know what data I am looking at. Also, will this work in async? I am thinking about what happens when there are 100K rows....we need to make sure this won't freeze up their session if possible.

fixed no headers

wilwong89 (approved):

kyle1morel (approved):

PR #277, opened 2025-03-05
Reorganizes views folder structure

kyle1morel:
# Description

Reorganizes views and views unit tests folder structure to match new routing.

## Types of changes

Documentation (non-breaking change with enhancements to documentation)

## Checklist


frontend/src/router/internalHousing.ts
```diff
@@ -65,7 +66,8 @@ const routes: Array<RouteRecordRaw> = [
               {
                 path: ':permitId',
                 name: RouteName.INT_HOUSING_PROJECT_PROPONENT_PERMIT,
-                component: () => import('@/views/permit/PermitStatusView.vue'),
+                // TODO: Do we want to make an internal view that uses the same components?
```
qhanson55, line 69: didn't want to go file duplicate route? Also should have TODO reflect that we do want to eventually go down this route rather than a question?

kyle1morel: There's a significant amount of logic in the views themselves. I don't think duplicating all that is a good idea. It could potentially all be moved into a component? Then the views could just use that component. Kind of defeats the purpose of the view at that point it feels like, as it's effectively just acting solely as a wrapper. Idk. Ill update the comment for now.

qhanson55 (approved):

jujaga (approved):

PR #279, opened 2025-03-07
Fixed incorrect enquiry and submission search for contact history

qhanson55:
# Description

Fixed incorrect enquiry and submission search for contact history so that contacts with no user attachment have correct history and so that any nav created submissions are still connected to the correct contact.

## Types of changes

## Checklist

### DIFF
--- a/frontend/src/components/contact/ContactHistoryList.vue
+++ b/frontend/src/components/contact/ContactHistoryList.vue
@@ -6,7 +6,6 @@ import { useRoute, useRouter } from 'vue-router';
 import { Spinner } from '@/components/layout';
 import { Column, DataTable } from '@/lib/primevue';
 import { RouteName } from '@/utils/enums/application';
-import { IntakeStatus } from '@/utils/enums/housing';
 import { formatDate } from '@/utils/formatters';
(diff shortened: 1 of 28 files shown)


sanjaytkbabu (approved):

kyle1morel (approved):

frontend/src/components/contact/ContactHistoryList.vue
```diff
@@ -189,10 +188,14 @@ onMounted(() => {
...
+      field="applicationStatus"
       header="Activity status"
       :sortable="true"
       style="min-width: 150px"
-    />
+    >
+      <template #body="{ data }">
+        {{ 'submissionId' in data ? data.applicationStatus : data.enquiryStatus }}
+      </template>
+    </Column>
```
kyle1morel, line 199: Sorting will not work with enquiry status here as the field will be undefined. I think the only way to get around this though is to create two separate columns.

qhanson55: All submissions will have a `submissionId` and if not a submission then it is an Enquiry which will have the enquiryStatus, no?

kyle1morel (approved):

PR #285, opened 2025-03-20
New login flow

sanjaytkbabu:
# Description
[PADS-461](https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-461)
OIDCLoginView updated with the new login UI
LoginView removed
login method updated with kc_idp_hint

## Checklist


Subin1Doo:
Looks great! Just a couple of changes,,
- Can we open the hyperlinks in a different window so it's easier for the users to come back to the login page once they create an ID/ account. 
- Can we make it so the login area is a fixed width and height? it's 460px by 772px in the design. 
- browser back button not working as Sharolyn mentioned
- can we change the background colour to BC blue plz
Thank you!

kyle1morel (approved):

PR #293, opened 2025-04-01
Chore/q4 security continuity updates

qhanson55:
# Description

## Checklist


slhurley:
- Navigator entered submissions should not show on proponents My Projects page (see F2E1C324).  They just cause errors.
- On the intake form, choose that it is a Registered business and then try to actually choose one from the list after entering a couple of characters...it won't let you.
- Everything else looks good to me!

qhanson55:
> 1. Navigator entered submissions should not show on proponents My Projects page (see F2E1C324).  They just cause errors.
> 2. On the intake form, choose that it is a Registered business and then try to actually choose one from the list after entering a couple of characters...it won't let you.
> 3. Everything else looks good to me!

1. This behaviour replicates on our current master branch, but it is expected. This will only happen to users who have developer permissions and are able to both view Proponent Housing View and Navigator Housing View. When a developer creates a submission on the Nav side it will show under their projects on the Prop side. Outside of users with developer permissions you won't find Projects created by Navs in the Props' Project list as that list is scoped to self.
2. This is a known bug (but need a ticket)... it's intermittent and we can't get it to consistently replicate, is this consistently happening to you? Or does it not work sometimes and then work?
3. Thank you :)

jujaga (approved):

kyle1morel (approved):

PR #298, opened 2025-04-15
API split for initiatives

kyle1morel:
# Description

- Adds a new `PCNS` global initiative, certain routes still reside at this top level under `api/v1`
  - `/ats`, `/contact`, `/docs`,  `/reporting`, `/sso`, `/user`, `/yars`
- Electrification and Housing specific routes are now nested under `api/v1/electrification` and `api/v1/housing`
  - `/accessRequest`, `/document`, `/enquiry`, `/project`, `/map`, `/note`, `/permit`, `/roadmap`
- Front end service layers for electrification and housing services now rely on the active initiative from the app store to determine correct route
- `YARS` resource added to protect certain RBAC endpoints
- Contact history now only shows submissions for the current initiative 

## Types of changes

## Checklist

### DIFF
--- a/frontend/src/components/contact/ContactHistoryList.vue
+++ b/frontend/src/components/contact/ContactHistoryList.vue
@@ -71,7 +71,7 @@ function normalizeContactHistory() {
 
 function updateQueryParams() {
   router.replace({
-    name: RouteName.INT_CONTACT_PAGE,
+    name: router.currentRoute.value.name,
     query: {
       rows: pagination.value.rows ?? undefined,
       order: pagination.value.order ?? undefined,
--- a/frontend/src/components/contact/ContactHistoryList.vue
+++ b/frontend/src/components/contact/ContactHistoryList.vue
@@ -10,12 +10,12 @@ import { formatDate } from '@/utils/formatters';
 import { toNumber } from '@/utils/utils';
 
 import type { Ref } from 'vue';
-import type { Enquiry, HousingProject, Pagination } from '@/types';
+import type { ElectrificationProject, Enquiry, HousingProject, Pagination } from '@/types';
 
 // Props
 const { assignedUsers, contactsHistory, loading } = defineProps<{
   assignedUsers: Record<string, string>;
-  contactsHistory: Array<HousingProject | Enquiry>;
+  contactsHistory: Array<ElectrificationProject | HousingProject | Enquiry>;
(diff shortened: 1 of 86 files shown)


slhurley:
Knowing I can't test everything yet...the UI looks good.  A couple of comments:
- Auto-save moves a little quick (maybe ok once we fix the bug in backlog)
- The menubar has "Housing" options...it shouldn't...if this is a common component we'll need to reword that.
- "Back to Electrification" on the confirmation page takes you to Housing.  Since we want to release Nav side before tracker perhaps it could be unlinked until we have all pieces in place?
- The link on the confirmation email that is to take you to your greyed out intake form filled in actually takes you to a blank Housing intake form.

slhurley:
- Download prod data on the Stats page doesn't work
- In Contacts on Electrification, when you click on "View" to see more about a contact it errors out with "Missing required param enquiryID".  Also I'm going to assume for now this is only Electrification Contacts and no Housing Contacts are showing up here.

frontend/src/components/contact/ContactHistoryList.vue
```diff
@@ -58,9 +63,9 @@ function getRouteToObject(data: Enquiry | HousingProject) {
 function normalizeContactHistory() {
   return contactsHistory.map((se) => ({
     ...se,
-    state: 'housingProjectId' in se ? se.applicationStatus : se.enquiryStatus,
+    state: 'projectId' in se ? se.applicationStatus : se.enquiryStatus,
     assignedUser: se.assignedUserId ? getUsersName(se.assignedUserId) : 'Unassigned',
-    historyType: 'housingProjectId' in se ? 'Project' : 'Enquiry'
+    historyType: 'projectId' in se ? 'Project' : 'Enquiry'
```
qhanson55, line 68: Could "future proof" and have the true case return '<Initiative> Project', even if just a small chance it of props being in both... But even then, it may be useful for Navs to see the Initiative that is in the prop's history without clicking into it.

frontend/src/services/electrificationProjectService.ts
```diff
@@ -1,16 +1,19 @@
 import { appAxios } from './interceptors';
 import { delimitEmails } from '@/utils/utils';
+import { Initiative } from '@/utils/enums/application';
```
qhanson55, line 3: alpha order

PR #318, opened 2025-05-22
Ensure electrification works for proponents

qhanson55:
# Description

Refactored enums/constants into `projectCommon.ts` folders
Adjusted backend to work with electrification for proponents.
Adjust frontend for proponents project page and tracker to work with electrification.
Navs can now see prop's electrification view/tracker.
So we can open up electrification for proponents.
[PADS-530](https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-530) and [PADS-542](https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-542)

## Types of changes

## Checklist


Subin1Doo:
Electrification and tracker looks good on my end. 
Just one minor change in **proponent view** for both Housing & Elec, can we change the styling of 'cancel' buttons in confirmation dialogues? 
It's the confirmation dialogue that pops up when you try to delete a draft submission.
Styling change: primary blue for button outline and text, white background. Thanks!

frontend/src/utils/enums/projectCommon.ts
```diff
@@ -25,6 +31,16 @@ export enum EnquirySubmittedMethod {
...
+export enum IntakeFormCategory {
+  BASIC = 'basic',
+  CONTACTS = 'contacts',
+  HOUSING = 'housing',
+  LOCATION = 'location',
+  PERMITS = 'permits',
+  APPLIED_PERMITS = 'appliedPermits',
+  INVESTIGATE_PERMIS = 'investigatePermits'
+}
+
```
kyle1morel, line 43: This is specific to the housing intake, not common.

qhanson55: revisiting this later

qhanson55:
> 1. Start an Electric intake form, choose a business name for sure before auto-save.  Let it auto-save, Exit without saving and the draft will be there.  Open the draft and try to Submit without changing the business name.  Validation on the business name that was previously valid fails.
**Fixed**
> 2. On the Project List page, can we change the Project Types from displaying as "IPP_SOLAR" and make them readable like "IPP Solar"
**Fixed**
> 3. On the Tracker it shows the Navigator as "S D" if there is no Assignee
**Fixed**
> 4. I think we are missing that little icon on the tracker that allows them to see their greyed out project submission
I believe this was missing cause you were viewing from you developer profile, will have to try as a proponent to be able to see it. (Only props allowed to view greyed out project)
**Fixed**
> 5. From the email that you get from Enquiries, if you click on the link you get an Axios 500.  (Projects are fine just Enquiry emails)
**Fixed**

PR #320, opened 2025-05-27
Add basic feature flag toggling

kyle1morel:
# Description

Add basic feature flag toggling. JSON file read in at app start. Key/values are passed via `/config` endpoint to front end. New feature store added. A computed getters should be added for feature flag pulling from the config store.

## Types of changes

## Checklist

### DIFF
--- a/frontend/src/store/featureStore.ts
+++ b/frontend/src/store/featureStore.ts
@@ -1,0 +1,19 @@
+import { defineStore } from 'pinia';
+import { computed } from 'vue';
+
+import useConfigStore from '@/store/configStore';
+import { isTruthy } from '@/utils/utils';
+
+export const useFeatureStore = defineStore('feature', () => {
+  // Getters
+  const getters = {
+    isPeachEnabled: computed(() => isTruthy(useConfigStore().getConfig.features.peach))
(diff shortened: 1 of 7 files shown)


qhanson55 (approved):
Looks good to me, if we start to utilize the feature flag more often and consistently and new strategy may need to be implemented. But for seldom use, like PEACH, this will be more than suffice.

frontend/src/store/featureStore.ts
```diff
@@ -0,0 +1,19 @@
+import { defineStore } from 'pinia';
+import { computed } from 'vue';
+
+import useConfigStore from '@/store/configStore';
+import { isTruthy } from '@/utils/utils';
+
+export const useFeatureStore = defineStore('feature', () => {
+  // Getters
+  const getters = {
+    isPeachEnabled: computed(() => isTruthy(useConfigStore().getConfig.features.peach))
```
qhanson55, line 10: Future consideration, if we start having many feature flags (not likely anytime soon) we can do a single getter like `hasFeature(SOME_FEATURE_KEY)` similar to the `canNavigate` in `authzStore`

PR #324, opened 2025-05-29
Allow edit and delete on non-User Contacts via Contacts List

qhanson55:
# Description

Added delete contact API call.
Added new yars role, `CONTACT_UPDATER`, that allows proponents and read only navs to update their contact info without giving them the ability to delete contacts that comes from the `CONTACT_EDITOR` role.
Implemented new frontend flow and components to match [figma](https://www.figma.com/design/uW80J51g8OHm5fWICDdW9x/PAT--Latest-?node-id=7254-17104&t=b82gto5vvSOJxnSK-1) for Navs to update and delete contacts.
So that Navs can manage contacts safely away from the context of projects and enquiries.
[PADS-535](https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-535)

## Types of changes

## Checklist

### DIFF
--- a/app/src/routes/v1/contact.ts
+++ b/app/src/routes/v1/contact.ts
@@ -57,4 +57,13 @@ router.put(
   }
 );
 
+/** Delete a specific contact */
+router.delete(
+  '/:contactId',
+  hasAuthorization(Resource.CONTACT, Action.DELETE),
+  (req: Request<{ contactId: string }>, res: Response, next: NextFunction): void => {
+    contactController.deleteContact(req, res, next);
+  }
+);
+
 export default router;
--- a/app/src/routes/v1/contact.ts
+++ b/app/src/routes/v1/contact.ts
@@ -61,6 +61,7 @@ router.put(
 router.delete(
   '/:contactId',
   hasAuthorization(Resource.CONTACT, Action.DELETE),
+  contactValidator.deleteContact,
   (req: Request<{ contactId: string }>, res: Response, next: NextFunction): void => {
     contactController.deleteContact(req, res, next);
   }
--- a/app/src/services/contact.ts
+++ b/app/src/services/contact.ts
@@ -5,6 +5,23 @@ import { Contact, ContactSearchParameters, CurrentContext } from '../types';
 
 const service = {
   /**
+   * @function deleteContact
+   * Deletes a specific contact from the PCNS database
+   * @param {string} contactId Contact ID
+   * @returns {Promise<boolean>} The result of running the deleteMany operation if count is greater than 0
+   */
+  deleteContact: async (contactId: string) => {
+    if (!contactId) throw new Error('Contact ID is required for deletion');
+    const response = await prisma.contact.deleteMany({
+      where: {
+        contact_id: contactId
+      }
+    });
+
+    return response.count > 0;
+  },
+
+  /**
    * @function getContact
    * Gets a specific contact from the PCNS database
    * @param {string} contactId Contact ID
--- a/app/src/services/contact.ts
+++ b/app/src/services/contact.ts
@@ -8,17 +8,9 @@ const service = {
    * @function deleteContact
    * Deletes a specific contact from the PCNS database
    * @param {string} contactId Contact ID
-   * @returns {Promise<boolean>} The result of running the deleteMany operation if count is greater than 0
    */
   deleteContact: async (contactId: string) => {
-    if (!contactId) throw new Error('Contact ID is required for deletion');
-    const response = await prisma.contact.deleteMany({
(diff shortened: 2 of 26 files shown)


slhurley:
Everything works great...except I am wondering about allowing deletion when they have a submission attached.  It leaves the submission in a weird state with no contact.  I am thinking we should only allow deletion if the contact is no longer associated with a submission and instead add a toast telling them to associate the submission with someone else before deleting the contact?  @Subin1Doo ?

Subin1Doo:
> Everything works great...except I am wondering about allowing deletion when they have a submission attached. It leaves the submission in a weird state with no contact. I am thinking we should only allow deletion if the contact is no longer associated with a submission and instead add a toast telling them to associate the submission with someone else before deleting the contact? @Subin1Doo ?

That's true,, if that's the case,, we should notify them beforehand and prevent them from trying to delete it. 
We should deactivate the delete buttons for contacts that are manually created but has one or more submissions attached. We can add a warning banner for those contacts - "This contact is linked to submissions and can't be deleted. To proceed, delete the submissions or update them with a different contact." @slhurley

Subin1Doo:
Oh forgot to mention, can we move down the 'save and cancel' buttons on the 'edit contact' modal by a little bit? The buttons are closer to the 'preferred contact method' field rn. Thanks!

app/src/routes/v1/contact.ts
```diff
@@ -57,4 +57,13 @@ router.put(
...
 
+/** Delete a specific contact */
+router.delete(
+  '/:contactId',
+  hasAuthorization(Resource.CONTACT, Action.DELETE),
+  (req: Request<{ contactId: string }>, res: Response, next: NextFunction): void => {
+    contactController.deleteContact(req, res, next);
+  }
+);
+
```
kyle1morel, line 68: Missing validator

app/src/validators/contact.ts
```diff
@@ -23,6 +23,11 @@ export const contacts = Joi.array()
   .allow(null);
 
 const schema = {
+  deleteContact: {
+    params: Joi.object({
+      contactId: uuidv4.required()
+    })
+  },
```
kyle1morel, line 30: Validator not being passed in default export

app/src/validators/contact.ts
```diff
@@ -52,7 +57,7 @@ const schema = {
   },
   updateContact: {
     body: Joi.object({
-      userId: uuidv4.required(),
+      userId: uuidv4.allow(null),
```
kyle1morel, line 60: Validator modified, but no other changes made to contact updates?

app/src/services/contact.ts
```diff
@@ -4,6 +4,23 @@ import { generateCreateStamps, generateUpdateStamps } from '../db/utils/utils';
...
+  deleteContact: async (contactId: string) => {
+    if (!contactId) throw new Error('Contact ID is required for deletion');
+    const response = await prisma.contact.deleteMany({
+      where: {
+        contact_id: contactId
+      }
+    });
+
+    return response.count > 0;
+  },
```
kyle1morel, line 22: Waffling over `deleteMany` vs `delete` as we only expect to remove a single instance

app/src/validators/contact.ts
qhanson55: This was because if this is required we cannot make updates to contacts that don't have a userId.

app/src/services/contact.ts
qhanson55: I went with deleteMany as it was a simple way to return boolean on the db response (true if deleted, false if not), rather than returning a straight boolean on some conditional. Not totally against going the other way. Your call.

kyle1morel: Would prefer a single delete. Assuming `delete` returns null or undefined if the entity isn't found, then you can do `return !!response;` to still get a boolean.

frontend/src/components/contact/ContactsProponentsList.vue
```diff
@@ -58,6 +106,17 @@ onBeforeMount(() => {
...
+  // Map out contact history for each contact
+  contacts?.forEach(async (contact) => {
+    const contactData = (await contactService.getContact(contact.contactId, true)).data;
+    console.log(contactData);
+    if (contactData.activityContact.length) {
+      historyMap.value[contact.contactId] = true;
+    } else {
+      historyMap.value[contact.contactId] = false;
+    }
+  });
```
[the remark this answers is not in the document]
qhanson55: Changed contact search API to include activities if `includeActivities` param is true so that we don't have to ping the db for every contact. Then build map off of the included `contact_activity` field

PR #334, opened 2025-07-16
Multi-tracking Authorization Edit

sanjaytkbabu:
# Description
[PADS-565](https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-565) - Multi-auth edit

Backend:
Permit route, controller, service updates - createPermit/updatePermit replaced with upsertPermit
Permit validator updated
Api spec updated
Tests updated

Frontend:
Authorization edit functionality added
Authorization form updated to use permit store to get permits and pass them to authorization components
en-CA updated
Unused service permitNoteService.updatePermit removed
Tests updated 

## Checklist

### DIFF
--- a/app/src/controllers/housingProject.ts
+++ b/app/src/controllers/housingProject.ts
@@ -240,8 +240,8 @@ const controller = {
       });
 
       // Create each permit
-      await Promise.all(appliedPermits.map(async (x: Permit) => await permitService.createPermit(x)));
-      await Promise.all(investigatePermits.map(async (x: Permit) => await permitService.createPermit(x)));
+      await Promise.all(appliedPermits.map(async (x: Permit) => await permitService.upsertPermit(x)));
+      await Promise.all(investigatePermits.map(async (x: Permit) => await permitService.upsertPermit(x)));
 
       res.status(201).json(result);
     } catch (e: unknown) {
@@ -393,8 +393,8 @@ const controller = {
       });
 
       // Create each permit
-      await Promise.all(appliedPermits.map((x: Permit) => permitService.createPermit(x)));
-      await Promise.all(investigatePermits.map((x: Permit) => permitService.createPermit(x)));
+      await Promise.all(appliedPermits.map((x: Permit) => permitService.upsertPermit(x)));
+      await Promise.all(investigatePermits.map((x: Permit) => permitService.upsertPermit(x)));
 
       // Delete old draft
       if (req.body.draftId) await draftService.deleteDraft(req.body.draftId);
--- a/app/src/controllers/housingProject.ts
+++ b/app/src/controllers/housingProject.ts
@@ -9,7 +9,8 @@ import {
   emailService,
   enquiryService,
   housingProjectService,
-  permitService
+  permitService,
+  permitTrackingService
 } from '../services';
 import { Initiative } from '../utils/enums/application';
 import { NumResidentialUnits } from '../utils/enums/housing';
@@ -240,8 +241,26 @@ const controller = {
       });
 
       // Create each permit
-      await Promise.all(appliedPermits.map(async (x: Permit) => await permitService.upsertPermit(x)));
-      await Promise.all(investigatePermits.map(async (x: Permit) => await permitService.upsertPermit(x)));
+      await Promise.all(
+        appliedPermits.map(async (x: Permit) => {
+          const permitDataWithId = {
+            ...x,
+            permitId: x.permitId || uuidv4()
+          };
+          await permitService.upsertPermit(permitDataWithId);
+          await permitTrackingService.upsertPermitTracking(permitDataWithId);
+        })
+      );
+      await Promise.all(
+        investigatePermits.map(async (x: Permit) => {
+          const permitDataWithId = {
+            ...x,
+            permitId: x.permitId || uuidv4()
+          };
+          await permitService.upsertPermit(permitDataWithId);
+          await permitTrackingService.upsertPermitTracking(permitDataWithId);
+        })
+      );
 
       res.status(201).json(result);
     } catch (e: unknown) {
@@ -393,8 +412,26 @@ const controller = {
       });
 
       // Create each permit
-      await Promise.all(appliedPermits.map((x: Permit) => permitService.upsertPermit(x)));
-      await Promise.all(investigatePermits.map((x: Permit) => permitService.upsertPermit(x)));
+      await Promise.all(
+        appliedPermits.map(async (x: Permit) => {
+          const permitDataWithId = {
+            ...x,
+            permitId: x.permitId || uuidv4()
+          };
+          await permitService.upsertPermit(permitDataWithId);
+          await permitTrackingService.upsertPermitTracking(permitDataWithId);
+        })
+      );
+      await Promise.all(
+        investigatePermits.map(async (x: Permit) => {
+          const permitDataWithId = {
+            ...x,
+            permitId: x.permitId || uuidv4()
+          };
+          await permitService.upsertPermit(permitDataWithId);
+          await permitTrackingService.upsertPermitTracking(permitDataWithId);
+        })
+      );
 
       // Delete old draft
       if (req.body.draftId) await draftService.deleteDraft(req.body.draftId);
--- a/app/src/docs/v1.api-spec.yaml
+++ b/app/src/docs/v1.api-spec.yaml
@@ -916,7 +916,7 @@ paths:
               $ref: '#/components/schemas/Request-Permit'
       responses:
         '200':
-          description: Permit upserted successfully
+          description: Permit updated successfully
           content:
             application/json:
               schema:
--- a/app/src/docs/v1.api-spec.yaml
+++ b/app/src/docs/v1.api-spec.yaml
@@ -873,32 +873,6 @@ paths:
           $ref: "#/components/responses/UnprocessableEntity"
         default:
           $ref: '#/components/responses/Error'
-    put:
-      summary: Create a permit
-      operationId: createElectrificationPermit
-      tags:
-        - Electrification Permit
-      requestBody:
-        required: true
-        content:
-          application/json:
-            schema:
-              $ref: '#/components/schemas/Request-Permit'
-      responses:
-        '201':
-          description: Permit created successfully
-          content:
-            application/json:
-              schema:
-                $ref: '#/components/schemas/DB-Permit'
-        "401":
-          $ref: "#/components/responses/Unauthorized"
-        '403':
-          $ref: '#/components/responses/Forbidden'
-        "422":
-          $ref: "#/components/responses/UnprocessableEntity"
-        default:
-          $ref: '#/components/responses/Error'
   /electrification/permit/{permitId}:
     get:
       summary: Get a permit
@@ -930,17 +904,10 @@ paths:
         default:
           $ref: '#/components/responses/Error'
     put:
-      summary: Update a permit
-      operationId: updateElectrificationPermit
+      summary: Upsert a permit
+      operationId: upsertElectrificationPermit
       tags:
         - Electrification Permit
-      parameters:
-        - in: path
-          name: permitId
-          required: true
-          schema:
-            type: string
-            description: The ID of the permit to update
       requestBody:
         required: true
         content:
@@ -949,7 +916,7 @@ paths:
               $ref: '#/components/schemas/Request-Permit'
       responses:
         '200':
-          description: Permit updated successfully
+          description: Permit upserted successfully
           content:
             application/json:
               schema:
@@ -1988,32 +1955,6 @@ paths:
           $ref: "#/components/responses/UnprocessableEntity"
         default:
           $ref: '#/components/responses/Error'
-    put:
-      summary: Create a permit
-      operationId: createHousingPermit
-      tags:
-        - Housing Permit
-      requestBody:
-        required: true
-        content:
-          application/json:
-            schema:
-              $ref: '#/components/schemas/Request-Permit'
-      responses:
-        '201':
-          description: Permit created successfully
-          content:
-            application/json:
-              schema:
-                $ref: '#/components/schemas/DB-Permit'
-        "401":
-          $ref: "#/components/responses/Unauthorized"
-        '403':
-          $ref: '#/components/responses/Forbidden'
-        "422":
-          $ref: "#/components/responses/UnprocessableEntity"
-        default:
-          $ref: '#/components/responses/Error'
   /housing/permit/{permitId}:
     get:
       summary: Get a permit
@@ -2029,7 +1970,7 @@ paths:
             description: The ID of the permit to update
       responses:
         '200':
-          description: Permit updated successfully
+          description: Permit upserted successfully
           content:
             application/json:
               schema:
@@ -2045,17 +1986,10 @@ paths:
         default:
           $ref: '#/components/responses/Error'
     put:
-      summary: Update a permit
-      operationId: updateHousingPermit
+      summary: Upsert a permit
+      operationId: upsertHousingPermit
       tags:
         - Housing Permit
-      parameters:
-        - in: path
-          name: permitId
-          required: true
-          schema:
-            type: string
-            description: The ID of the permit to update
       requestBody:
         required: true
         content:
--- a/app/src/services/permit.ts
+++ b/app/src/services/permit.ts
@@ -10,45 +10,6 @@ import type { ListPermitsOptions, Permit, PermitTracking } from '../types';
 
 const service = {
   /**
-   * @function createPermit
-   * Creates a Permit
-   * @param {Permit} data Permit object
-   * @returns {Promise<Permit | null>} The result of running the create operation
-   */
-  createPermit: async (data: Permit) => {
-    try {
-      const response = await prisma.$transaction(async (trx) => {
-        const newPermit = { ...data, permitId: uuidv4() };
-
-        const createResponse = await trx.permit.create({
-          include: {
-            permit_type: true
-          },
-          data: { ...permit.toPrismaModel(newPermit), created_by: data.createdBy, updated_by: data.updatedBy }
-        });
-
-        if (data?.permitTracking) {
-          await trx.permit_tracking.createMany({
-            data: data?.permitTracking?.map((x: PermitTracking) => ({
-              permitId: newPermit?.permitId,
-              trackingId: x.trackingId,
-              sourceSystemKindId: x?.sourceSystemKindId,
-              shownToProponent: x?.shownToProponent ?? false,
-              createdBy: data.createdBy,
-              updatedBy: data.updatedBy
-            }))
-          });
-        }
-        return createResponse;
-      });
-
-      return permit.fromPrismaModel(response);
-    } catch (e: unknown) {
-      throw e;
-    }
-  },
-
-  /**
    * @function deletePermit
    * Delete a permit
    * @param {string} permitId Permit ID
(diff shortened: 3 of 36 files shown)


slhurley:
- If you go to add an ID and try to save without touching the status update, it seems to think you need to update the note but you don't as the status has not changed.
- If you click in the breadcrumbs after making any change it does not tell you have not saved yet and loses the info.  Same if you use the back button. I think we need the "unsaved changes" warning.
- Now that we have turned this into a page, I feel the breadcrumbs need to allow them to go back to the Auth tab?
- I don't think you should be able to select the exact same tracking id more than once.

naixin-zhangbc:
- After a new authorization is added, the tracking IDs of the existing authorization get removed somehow. 
- When trying to change the "Needed" field from "Under investigation" to "Yes" or "No", it shows "Error: Failed to save permit. Request failed with status code 422"
- Not part of this ticket... just wanted to mention in case it's related to other issues... the Submitted Date doesn't show in the notification email when an update note is added to an authorization.
- Please update the grey background for the "agency, business domain..." row and the "update note history" to FAF9F8

sanjaytkbabu:
> * After a new authorization is added, the tracking IDs of the existing authorizations get removed somehow.

fixed!
> * When trying to change the "Needed" field from "Under investigation" to "Yes" or "No", it shows "Error: Failed to save permit. Request failed with status code 422"

It was caused by missing status verification date, it's been fixed as well.
> * Not part of this ticket... mentioning it in case it's related to other issues... the Submitted Date doesn't show up in the notification email. The email that's sent to Prop when an update note is added to an authorization.

Thank you for mentioning this, it happened due to date being returned in a different format. Fixed!
> * Wondering if it's possible to show the "Authorization type and IDs" section expanded, when the authorization come in through Props' intake form and when it's the first time Navs open it up. Asking because that's when Navs need to use that section to add in Tracking IDs.

Will get back to you on this.
> * Please update the grey background for the "agency, business domain..." row and the "update note history" to FAF9F8

done!

sanjaytkbabu:
> * If you go to add an ID and try to save without touching the status update, it seems to think you need to update the note but you don't as the status has not changed.

done!
> * If you click in the breadcrumbs after making any change it does not tell you have not saved yet and loses the info.  Same if you use the back button. I think we need the "unsaved changes" warning.
> * Now that we have turned this into a page, I feel the breadcrumbs need to allow them to go back to the Auth tab?
> * I don't think you should be able to select the exact same tracking id more than once.

slhurley:
- In the bottom where Business Domain and Agency is shown, can we have a space between the label and the value?
- One thing that is a bit of a problem is when the prop fills out the tracking id on the intake form it is not linked to any particular tracking id name when it is saved.  I guess this is because we have not linked names to particular permits.  We will need to think about how to handle this in future.

Subin1Doo:
Can we also vertically left align "Agency" with the "Update note history" header while keeping the spacing equally distributed between Agency, business domain, updated by, and last updated.

Subin1Doo:
the date and time format needs commas according to the BC guidelines, we should stick to this from now on,, so format should be [December 31, 2016] for just the date, date and time [September 25, 2015, 1:30 pm]. Thanks!

Subin1Doo:
Screenshot 2025-07-18 at 11 53 56 AM
We need some spacing between the 'update' text section and permit numbers box.

sanjaytkbabu:
> * In the bottom where Business Domain and Agency is shown, can we have a space between the label and the value?

done!
> * One thing that is a bit of a problem is when the prop fills out the tracking id on the intake form it is not linked to any particular tracking id name when it is saved.  I guess this is because we have not linked names to particular permits.  We will need to think about how to handle this in future.

app/src/controllers/housingProject.ts
```diff
@@ -240,8 +240,8 @@ const controller = {
       });
 
       // Create each permit
-      await Promise.all(appliedPermits.map(async (x: Permit) => await permitService.createPermit(x)));
-      await Promise.all(investigatePermits.map(async (x: Permit) => await permitService.createPermit(x)));
+      await Promise.all(appliedPermits.map(async (x: Permit) => await permitService.upsertPermit(x)));
+      await Promise.all(investigatePermits.map(async (x: Permit) => await permitService.upsertPermit(x)));
```
jujaga, line 244: The map callback function does not need an async and await - the permitService should be a Promise function already and should be handled by Promise.all's collection.

app/src/docs/v1.api-spec.yaml
```diff
@@ -949,7 +916,7 @@ paths:
               $ref: '#/components/schemas/Request-Permit'
       responses:
         '200':
-          description: Permit updated successfully
+          description: Permit upserted successfully
```
jujaga, line 919: Will be something to converse about, but generally API documentation focuses on the outcome of the action, not how it was implemented. In this case, it may make sense to retain the updated instead of changing to upserted, because it is fundamentally an update, only that you achieve it with an upsert operation instead.

app/src/services/permit.ts
```diff
@@ -173,22 +134,79 @@ const service = {
...
+                    shownToProponent: x?.shownToProponent ?? false,
+                    createdBy: permitDataWithId.createdBy,
+                    updatedBy: permitDataWithId.updatedBy
+                  }
+                });
+              }
+            })
+          );
         }
+        return response;
```
jujaga, line 207: Two observations:
- The three main blocks of operations may be await Promise.all wrapped together - I'm not seeing any immediate ordering constraints on the three operations (aka something must execute first as the next call is dependent on the result)
- The result of trx.permit.upsert is only used on the return - may be good idea to do this operation last, as then you could just return that action instead of declaring another const just to use it in one place.
There's an opportunity to streamline the async subfunctions here.

app/src/services/permit.ts
```diff
@@ -173,22 +134,79 @@ const service = {
...
+        });
+
+        // Delete any permit tracking that is not in the new data
+        await trx.permit_tracking.deleteMany({
+          where: {
+            permitId: permitDataWithId.permitId,
+            permitTrackingId: {
+              notIn: permitDataWithId.permitTracking
+                ?.map((x: PermitTracking) => x.permitTrackingId)
+                .filter((x) => x) as number[]
```
jujaga, line 170: If possible, we should try to avoid using Typescript as assertions, as they should be considered tools of last resort. You should be able to cast this earlier in the chain as a number like this, which properly asserts the return type of map and cascades further down:

```ts
?.map<number>((x: PermitTracking) => x.permitTrackingId)
.filter((x) => x)
```

frontend/src/components/authorization/AuthorizationForm.vue
```diff
@@ -1,67 +1,72 @@
...
 // Store
 const { getConfig } = storeToRefs(useConfigStore());
+const { getProject, getPermits } = storeToRefs(projectStore);
+
+// State
+const authorization: Ref<Permit | undefined> = ref(undefined);
+const formRef: Ref<InstanceType<typeof Form> | null> = ref(null);
+const initialFormValues: Ref<any | undefined> = ref(undefined);
+const authorizationNotes: Ref<Array<PermitNote> | undefined> = ref([]);
+const updatedBy: Ref<User | undefined> = ref(undefined);
```
jujaga, line 50: Alphabetical ordering

frontend/src/components/authorization/AuthorizationForm.vue
```diff
@@ -155,6 +190,44 @@ const formSchema = object({
...
+function initializeFormValues() {
+  if (authorizationId) {
+    authorization.value = getPermits.value.find((p) => p.permitId === authorizationId) as Permit;
+    authorizationNotes.value = authorization.value.permitNote || [];
+    if (authorization.value) {
+      initialFormValues.value = {
+        authorizationType: authorization.value.permitType,
+        adjudicationDate: authorization.value.adjudicationDate ? new Date(authorization.value.adjudicationDate) : null,
+        submittedDate: authorization.value.submittedDate ? new Date(authorization.value.submittedDate) : null,
+        /* eslint-disable indent */
```
jujaga, line 203: We should not be overriding eslint like this without a very good reason otherwise. I don't see a companion enable clause either.

frontend/tests/unit/service/permitService.spec.ts
```diff
@@ -59,7 +59,7 @@ vi.mocked(appAxios).mockReturnValue({
 } as any);
 
 // Spies
-const updatePermitSpy = vi.spyOn(permitService, 'updatePermit');
+const upsertPermitSpy = vi.spyOn(permitService, 'upsertPermit');
```
jujaga, line 62: Please make sure to address any test and lint warnings as you see them.

app/src/services/permit.ts
```diff
@@ -173,22 +134,79 @@ const service = {
...
+          }
+        });
+
+        // Upsert the permit tracking
+        if (permitDataWithId.permitTracking?.length) {
+          await Promise.all(
+            permitDataWithId.permitTracking.map(async (x: PermitTracking) => {
+              if (x.permitTrackingId) {
+                await trx.permit_tracking.update({
+                  where: {
```
kyle1morel, line 181: Upsert calls for `permit_tracking` table does not belong in `permit` service.

frontend/src/components/authorization/AuthorizationCard.vue
```diff
@@ -81,30 +90,50 @@ watchEffect(() => {
...
+                v-tooltip.right="t('authorization.authorizationCard.clickToCopy')"
+                class="cursor-pointer"
+                @click="toCopy(trackingShownToProponent?.trackingId as string)"
+              >
+                <span class="font-bold">
+                  {{ trackingShownToProponent?.sourceSystemKind?.description }}
+                </span>
+                :
+
+                {{ trackingShownToProponent?.trackingId }}
```
kyle1morel, line 103: `?` should be unnecessary after `trackingShownToProponent` as this whole thing is warpped in a `v-if` on L86.

frontend/src/components/authorization/AuthorizationCard.vue
```diff
@@ -81,30 +90,50 @@ watchEffect(() => {
...
-              {{ permitTracking?.trackingId }}
+              <span
+                v-tooltip.right="t('authorization.authorizationCard.clickToCopy')"
+                class="cursor-pointer"
+                @click="toCopy(permitTracking?.trackingId as string)"
+              >
+                <span class="font-bold">{{ permitTracking.sourceSystemKind?.description }}</span>
+                :
+                {{ permitTracking?.trackingId }}
+              </span>
```
kyle1morel, line 119: Unnecessary `?` after `permitTracking`

frontend/src/components/authorization/AuthorizationCard.vue
```diff
@@ -81,30 +90,50 @@ watchEffect(() => {
...
               :key="permitTracking.permitTrackingId"
               class="ml-5 mt-2 flex items-center"
             >
-              <span class="font-bold">{{ permitTracking.sourceSystemKind?.description }}</span>
-              :
-              {{ permitTracking?.trackingId }}
+              <span
+                v-tooltip.right="t('authorization.authorizationCard.clickToCopy')"
+                class="cursor-pointer"
+                @click="toCopy(permitTracking?.trackingId as string)"
```
kyle1morel, line 114: Unnecessary casting. `permitTracking` is guaranteed to exist so the `?` can be removed.

frontend/src/components/authorization/AuthorizationCardLite.vue
```diff
@@ -15,6 +15,9 @@ const { permit } = defineProps<{
   permit: Permit;
 }>();
 
+// Emits
+const emit = defineEmits(['authorizationCard:more']);
```
kyle1morel, line 19: Emit name does not match file.

frontend/src/components/authorization/AuthorizationForm.vue
```diff
@@ -1,67 +1,72 @@
...
 
 // Store
 const { getConfig } = storeToRefs(useConfigStore());
+const { getProject, getPermits } = storeToRefs(projectStore);
+
+// State
+const authorization: Ref<Permit | undefined> = ref(undefined);
+const formRef: Ref<InstanceType<typeof Form> | null> = ref(null);
+const initialFormValues: Ref<any | undefined> = ref(undefined);
+const authorizationNotes: Ref<Array<PermitNote> | undefined> = ref([]);
```
kyle1morel, line 49: Arrays should not be possibly undefined. Instead favour setting them to an empty array. This prevents `?` checking before calling array functions.

frontend/src/components/authorization/AuthorizationForm.vue
kyle1morel: If this is a prettier formatting issue use the `prettier-ignore` directive instead.

frontend/src/components/authorization/AuthorizationForm.vue
```diff
@@ -187,39 +267,63 @@ const formSchema = object({
...
+          icon="pi pi-check"
+        />
+        <Button
+          class="p-button-outlined mr-2"
+          :label="t('authorization.authorizationForm.cancel')"
+          icon="pi pi-times"
+          @click="
+            router.push({
+              name: projectRouteName,
+              params: { projectId: projectId },
```
kyle1morel, line 285: I think you can use `params: { projectId },` here. Shouldn't need to redeclare it as the same name.

frontend/src/components/authorization/AuthorizationForm.vue
sanjaytkbabu: tried prettier-ignore, but got error `Expected indentation of 12 spaces but found 14.eslint[indent](https://eslint.org/docs/latest/rules/indent)`

app/src/services/permit.ts
```diff
@@ -173,22 +134,79 @@ const service = {
...
+  upsertPermit: async (data: Permit) => {
     try {
-      const response = await prisma.permit.update({
-        include: {
-          permit_type: true
-        },
-        data: { ...permit.toPrismaModel(data), updated_by: data.updatedBy },
-        where: {
-          permit_id: data.permitId
+      const permitDataWithId = { ...data, permitId: data.permitId || uuidv4() };
```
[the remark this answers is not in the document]
sanjaytkbabu: moved the logic to controller layer.

frontend/src/components/authorization/AuthorizationForm.vue
sanjaytkbabu: this works, I removed it from 3 places in the same file.

frontend/tests/unit/service/permitService.spec.ts
sanjaytkbabu: Had commented out a test to deal with it later. I just fixed it!

app/src/services/permit.ts
jujaga: In that case, you need to do one of the following:
- Declare that x.permitTrackingId is not going to be a null value with `x.permitTrackingId!` with trailing exclamation
- `?.map<number | undefined>((x: PermitTracking) => x.permitTrackingId)`

sanjaytkbabu:
> * If you click in the breadcrumbs after making any change it does not tell you have not saved yet and loses the info.  Same if you use the back button. I think we need the "unsaved changes" warning.

done!
> * Now that we have turned this into a page, I feel the breadcrumbs need to allow them to go back to the Auth tab?
> * I don't think you should be able to select the exact same tracking id more than once.

I spoke with Kyle, Because of the way we've setup breadcrumbs and tracking information rows being dynamic it'd too difficult and the effort wold be much more than what it's worth.

naixin-zhangbc:
> > On the Add/Edit an authorization page
> > 
> > * When a required field is left empty upon publishing, should it trigger an alert toast, or scroll to the required field? Whichever is used elsewhere on Nav side would be good. I see in the Prop intake form, it takes users to the missing fields.
> 
> done!
> 
> > * When the sections are collapsed (Authorization types and IDs, Status updates, etc.), can we make the entire header row clickable to expand the section? So that users don't have to click on the +/- button precisely.
> 
> In talks with Sharolyn, we'd have to do this change across the board to keep things consistent (electrification). We're essentially changing from Panel to Accordion component. Concluded that the effort is more than what it's worth for this ticket.
> 
> > * Thought I saw this been fixed before that by default Authorization Status should show "None", and Application Stage "Pre-submission". When adding a new authorization, these two fields are blank rn.
> 
> Oops, it was missed when we did the edit, put it back!
> 
> > * This is how the business domain row shows up when opening the Add authorization page. Can we get rid of the empty space and comma?
> 
> Removed it on add auth, only appears during edit.

Thank you!

Here's the new issue we discovered

- When adding a new authorization, upon publishing, it shows this message, clears all the filled info, and shows the green success toast in the background (didn't capture it in the screenshot). This seems to only happen when a tracking id field is added.
Screenshot 2025-07-23 at 9 37 11 AM

sanjaytkbabu:
> > > On the Add/Edit an authorization page
> > > 
> > > * When a required field is left empty upon publishing, should it trigger an alert toast, or scroll to the required field? Whichever is used elsewhere on Nav side would be good. I see in the Prop intake form, it takes users to the missing fields.
> > 
> > 
> > done!
> > > * When the sections are collapsed (Authorization types and IDs, Status updates, etc.), can we make the entire header row clickable to expand the section? So that users don't have to click on the +/- button precisely.
> > 
> > 
> > In talks with Sharolyn, we'd have to do this change across the board to keep things consistent (electrification). We're essentially changing from Panel to Accordion component. Concluded that the effort is more than what it's worth for this ticket.
> > > * Thought I saw this been fixed before that by default Authorization Status should show "None", and Application Stage "Pre-submission". When adding a new authorization, these two fields are blank rn.
> > 
> > 
> > Oops, it was missed when we did the edit, put it back!
> > > * This is how the business domain row shows up when opening the Add authorization page. Can we get rid of the empty space and comma?
> > 
> > 
> > Removed it on add auth, only appears during edit.
> 
> Thank you!
> 
> Here's the new issue we discovered
> 
> * When adding a new authorization, upon publishing, it shows this message, clears all the filled info, and shows the green success toast in the background (didn't capture it in the screenshot). This seems to only happen when a tracking id field is added.
> 
> Screenshot 2025-07-23 at 9 37 11 AM

must be fixed now

app/tests/unit/controllers/permit.spec.ts
```diff
@@ -109,11 +112,12 @@ describe('createPermit', () => {
     });
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    await permitController.createPermit(req as any, res as any, next);
+    await permitController.upsertPermit(req as any, res as any, next);
 
     expect(createSpy).toHaveBeenCalledTimes(1);
     expect(createSpy).toHaveBeenCalledWith({
       ...req.body,
+      permitId: expect.any(String),
```
kyle1morel, line 120: `expect.stringMatching(uuidv4Pattern)`

app/tests/unit/controllers/permit.spec.ts
```diff
@@ -393,11 +397,12 @@ describe('updatePermit', () => {
     updateSpy.mockResolvedValue(updated);
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    await permitController.updatePermit(req as any, res as any, next);
+    await permitController.upsertPermit(req as any, res as any, next);
 
     expect(updateSpy).toHaveBeenCalledTimes(1);
     expect(updateSpy).toHaveBeenCalledWith({
       ...req.body,
+      permitId: expect.any(String),
```
kyle1morel, line 405: `expect.stringMatching(uuidv4Pattern)`

app/tests/unit/controllers/permit.spec.ts
```diff
@@ -428,11 +433,12 @@ describe('updatePermit', () => {
     });
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    await permitController.updatePermit(req as any, res as any, next);
+    await permitController.upsertPermit(req as any, res as any, next);
 
     expect(updateSpy).toHaveBeenCalledTimes(1);
     expect(updateSpy).toHaveBeenCalledWith({
       ...req.body,
+      permitId: expect.any(String),
```
kyle1morel, line 441: `expect.stringMatching(uuidv4Pattern)`

frontend/src/components/authorization/AuthorizationCard.vue
```diff
@@ -36,6 +40,11 @@ watchEffect(() => {
       .catch(() => {});
   }
 });
+
+function toCopy(toCopy: string) {
+  navigator.clipboard.writeText(toCopy);
+  toast.info(t('authorization.authorizationCard.copiedToClipboard'));
+}
```
kyle1morel, line 47: Style: Should be before lifecycle functions

frontend/src/components/authorization/AuthorizationCard.vue
```diff
@@ -81,30 +90,50 @@ watchEffect(() => {
...
         </div>
       </div>
     </template>
   </Card>
 </template>
+
+<style lang="scss" scoped>
+.cursor-pointer {
+  cursor: pointer;
+}
```
kyle1morel, line 138: Should exist in tailwind somewhere

frontend/src/components/authorization/AuthorizationForm.vue
```diff
@@ -1,67 +1,77 @@
 <script setup lang="ts">
 import { storeToRefs } from 'pinia';
 import { Form } from 'vee-validate';
-import { inject } from 'vue';
+import { inject, onBeforeMount, ref } from 'vue';
```
kyle1morel, line 4: Alphabetize

frontend/src/components/authorization/AuthorizationForm.vue
```diff
@@ -155,6 +195,66 @@ const formSchema = object({
...
+    let el = document.querySelector(`[name="${error}"]`);
+    let rect = el?.getBoundingClientRect();
+
+    if (rect) {
+      if (!first) first = el;
+      else if (rect.top < first.getBoundingClientRect().top) first = el;
+    }
+  }
+
+  first?.scrollIntoView({ behavior: 'smooth' });
```
kyle1morel, line 249: This bit is used across all our forms. We should turn this into a util function.

frontend/src/utils/utils.ts
```diff
@@ -274,6 +274,30 @@ export function partition<T>(
   );
 }
 
+/**
+ * @function scrollToFirstError
+ * Scrolls to the first invalid field in a form
+ * @param {object} e The event object containing validation errors
+ */
+export function scrollToFirstError(e: any) {
```
kyle1morel, line 282: `any` type should not be used. I believe the errors object is a simple object containing key/values. 

`export function scrollToFirstError(errors: { [key: string]: string })`

We can then also remove line 283, and when this function is used explicitly pass in only the errors instead of the entire event - `scrollToFirstError(e.errors);`

PR #353, opened 2025-08-21
Navigator Note Card Update

sanjaytkbabu:
# Description
[PADS-577](https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-577)

Navigator note card updates

## Checklist

### DIFF
--- a/frontend/src/components/authorization/AuthorizationCard.vue
+++ b/frontend/src/components/authorization/AuthorizationCard.vue
@@ -51,13 +51,17 @@ function toCopy(toCopy: string) {
     </template>
     <template #content>
       <div class="flex gap-2">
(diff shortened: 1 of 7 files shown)


Subin1Doo:
The icon and the text colour for 'shown to proponent' label should be the gov primary blue. Can we increase the corner radius to 4px for both labels? 

For the title hyperlink, can we make it behave the same as Authorization cards? The underline for title should show when user hovers over the title area. 

And for the vertical bars between the title, bf value, and escalation type, can we make them a big shorter in height so it's somewhat similar to the text height and darken the colour a little bit? 

Thanks!

Subin1Doo:
just realized we're missing the bf date for bf notes! It should come after the last updated date. Everything else looks good, thanks!

kyle1morel (approved):

PR #354, opened 2025-08-21
Request Sanitizing Middleware

qhanson55:
# Description

Implemented middleware that will sanitize string inputs to remove any unwanted null characters in case users copy and past them into inputs.

## Types of changes

## Checklist

## Further comments

Also added some utility function testings and change to ts docs


app/src/middleware/requestSanitizer.ts
```diff
@@ -0,0 +1,50 @@
...
+  // eslint-disable-next-line no-control-regex
+  return str.replace(/\u0000/g, '');
+}
+
+/**
+ * Recursively sanitize JSON-like data (strings, arrays, plain objects)
+ * @param value The value to sanitize
+ * @returns The sanitized value if a string otherwise the original value
+ */
+function sanitize<T>(value: T): T | string {
```
jujaga, line 18: These are utility functions that can fit in many places - may be worth housing them in `/src/utils/utils.ts` or equivalent instead.

app/src/middleware/requestSanitizer.ts
```diff
@@ -0,0 +1,50 @@
+import type { NextFunction, Request, Response } from 'express';
+
+/**
+ * Removes all NUL characters from a string
+ * @param str The string to sanitize
+ * @returns The sanitized string if it includes NUL character(s), otherwise the original string
+ */
+function stripNullCharactersFromString(str: string): string {
+  // eslint-disable-next-line no-control-regex
+  return str.replace(/\u0000/g, '');
```
jujaga, line 10: The eslint ignore should be one-lined when possible, similar to the no-console ones used periodically
```ts
return str.replace(/\u0000/g, ''); // eslint-disable-line no-control-regex
```

app/tests/unit/middleware/requestSanitizer.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+  return app;
+}
+
+describe('requestSanitizer middleware', () => {
+  let app: express.Express;
+  beforeEach(() => {
+    app = buildApp();
+  });
+
+  test('removes NUL written as \\0 (octal) inside JSON body', async () => {
```
jujaga, line 31: Style: While functionally equivalent, we should be using it(...) notation instead of test(...) notation.

app/src/middleware/requestSanitizer.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+/**
+ * Recursively sanitize JSON-like data (strings, arrays, plain objects)
+ * @param value The value to sanitize
+ * @returns The sanitized value if a string otherwise the original value
+ */
+function sanitize<T>(value: T): T | string {
+  // Strip NUL characters from strings
+  if (typeof value === 'string') return stripNullCharactersFromString(value);
```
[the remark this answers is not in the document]
jujaga: This one might be a decent idea, assuming it actually works 🤔

sanjaytkbabu (approved):

app/src/middleware/requestSanitizer.ts
qhanson55: I was thinking the same but then since this everything coming into our API will hit these middleware I don't see us using these functions anywhere else.

kyle1morel (approved):

PR #452, opened 2026-04-09
feat: search authorizations

sanjaytkbabu:
# Description
[PADS-766](https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-766) 

Added search authorizations with pagination api
Added search authorizations functionality to the frontend
Refactored localization file and the places where it is used.

## Checklist

### DIFF
--- a/app/src/controllers/permit.ts
+++ b/app/src/controllers/permit.ts
@@ -134,16 +134,12 @@ export const searchPermitsController = async (
 ) => {
   const response = await transactionWrapper<{ permits: Permit[]; totalRecords: number }>(
     async (tx: PrismaTransactionClient) => {
-      const options: SearchPermitsOptions = {
-        ...req.query
-      };
-
       // Validate it's not PCNS
       if (req.currentContext.initiative === Initiative.PCNS) {
         throw new Problem(400, { detail: 'Invalid initiative' });
       }
 
-      return await searchPermitsPaginated(tx, req.currentContext.initiative!, options);
+      return await searchPermitsPaginated(tx, req.currentContext.initiative!, req.query); // nosonar
     }
   );
   res.status(200).json(response);
--- a/app/src/controllers/permit.ts
+++ b/app/src/controllers/permit.ts
@@ -7,7 +7,14 @@ import { generateCreateStamps, generateUpdateStamps } from '../db/utils/utils.ts
 import { summarizePeachRecord } from '../parsers/peach.ts';
 import { email } from '../services/email.ts';
 import { getPeachRecord } from '../services/peach.ts';
-import { deletePermit, getPermit, getPermitTypes, listPermits, upsertPermit } from '../services/permit.ts';
+import {
+  deletePermit,
+  getPermit,
+  getPermitTypes,
+  listPermits,
+  searchPermitsPaginated,
+  upsertPermit
+} from '../services/permit.ts';
 import { createPermitNote } from '../services/permitNote.ts';
 import { deleteManyPermitTracking, upsertPermitTracking } from '../services/permitTracking.ts';
 import { getProjectByActivityId } from '../services/project.ts';
@@ -33,6 +40,7 @@ import type {
   PermitTracking,
   PermitType,
   PermitUpdateEmailParams,
+  SearchPermitsOptions,
   SourceSystemKind
 } from '../types/index.ts';
 
@@ -111,6 +119,34 @@ export const sendPermitUpdateEmail = async (params: PermitUpdateEmailParams) => 
   };
 
   await email(emailData);
+};
+
+/**
+ * Searches for permits based on provided query parameters
+ * @param req - Express request containing search options in query parameters
+ * @param res - Express response object
+ * @returns Promise resolving to permits array and total record count
+ * @throws {Problem} 400 error if initiative is PCNS (invalid for this search)
+ */
+export const searchPermitsController = async (
+  req: Request<never, never, never, SearchPermitsOptions>,
+  res: Response
+) => {
+  const response = await transactionWrapper<{ permits: Permit[]; totalRecords: number }>(
+    async (tx: PrismaTransactionClient) => {
+      const options: SearchPermitsOptions = {
+        ...req.query
+      };
+
+      // Validate it's not PCNS
+      if (req.currentContext.initiative === Initiative.PCNS) {
+        throw new Problem(400, { detail: 'Invalid initiative' });
+      }
+
+      return await searchPermitsPaginated(tx, req.currentContext.initiative!, options);
+    }
+  );
+  res.status(200).json(response);
 };
 
 /**
--- a/app/src/services/permit.ts
+++ b/app/src/services/permit.ts
@@ -187,8 +187,9 @@ export const searchPermitsPaginated = async (
 ): Promise<{ permits: Permit[]; totalRecords: number }> => {
   // Determine project table based on initiative, exclude PCNS
   const projectTableMap: Record<Exclude<Initiative, Initiative.PCNS>, string> = {
-    [Initiative.HOUSING]: 'housingProject',
-    [Initiative.ELECTRIFICATION]: 'electrificationProject'
+    [Initiative.ELECTRIFICATION]: 'electrificationProject',
+    [Initiative.GENERAL]: 'generalProject',
+    [Initiative.HOUSING]: 'housingProject'
   };
 
   const projectTable = projectTableMap[initiative];
--- a/app/src/services/permit.ts
+++ b/app/src/services/permit.ts
@@ -193,13 +193,12 @@ export const searchPermitsPaginated = async (
 
   const projectTable = projectTableMap[initiative];
 
-  const sortDirection = options?.sortOrder === '1' ? 'asc' : 'desc';
   const validSortFields = ['decisionDate', 'stage', 'state', 'statusLastChanged', 'submittedDate'];
 
-  // Default sorting
-  let orderBy: Record<string, 'asc' | 'desc'> = { submittedDate: 'desc' };
+  let orderBy: Record<string, 'asc' | 'desc'> | undefined;
 
-  if (options?.sortField && validSortFields.includes(options.sortField)) {
+  if (options?.sortOrder !== '0' && options?.sortField && validSortFields.includes(options.sortField)) {
+    const sortDirection = options.sortOrder === '1' ? 'asc' : 'desc';
     orderBy = { [options.sortField]: sortDirection };
   }
 
@@ -247,13 +246,13 @@ export const searchPermitsPaginated = async (
                         { projectName: { contains: options.searchTag, mode: 'insensitive' as const } },
                         { companyNameRegistered: { contains: options.searchTag, mode: 'insensitive' as const } },
                         // Only include location fields for initiatives that have them (not ELECTRIFICATION)
-                        ...(initiative !== Initiative.ELECTRIFICATION
-                          ? [
+                        ...(initiative === Initiative.ELECTRIFICATION
+                          ? []
+                          : [
                               { streetAddress: { contains: options.searchTag, mode: 'insensitive' as const } },
                               { locality: { contains: options.searchTag, mode: 'insensitive' as const } },
                               { province: { contains: options.searchTag, mode: 'insensitive' as const } }
-                            ]
-                          : [])
+                            ])
                       ]
                     }
                   }
--- a/app/src/services/permit.ts
+++ b/app/src/services/permit.ts
@@ -1,7 +1,14 @@
 import { Initiative } from '../utils/enums/application.ts';
 
 import type { PrismaTransactionClient } from '../db/dataConnection.ts';
-import type { ListPermitsOptions, Permit, PermitBase, PermitSearchParams, PermitType } from '../types/index.ts';
+import type {
+  ListPermitsOptions,
+  Permit,
+  PermitBase,
+  PermitSearchParams,
+  PermitType,
+  SearchPermitsOptions
+} from '../types/index.ts';
 
 /**
  * Delete a specific permit
@@ -164,6 +171,147 @@ export const searchPermits = async (tx: PrismaTransactionClient, params: PermitS
     }
   });
   return response;
+};
+
+/**
+ * Search and retrieve permits with pagination, filtering, and sorting
+ * @param tx Prisma transaction client
+ * @param initiative Initiative code (excludes PCNS)
+ * @param options Search and filter options
+ * @returns A Promise that resolves to an object with permits array and total count
+ */
+export const searchPermitsPaginated = async (
+  tx: PrismaTransactionClient,
+  initiative: Exclude<Initiative, Initiative.PCNS>,
+  options: SearchPermitsOptions
+): Promise<{ permits: Permit[]; totalRecords: number }> => {
+  // Determine project table based on initiative, exclude PCNS
+  const projectTableMap: Record<Exclude<Initiative, Initiative.PCNS>, string> = {
+    [Initiative.HOUSING]: 'housingProject',
+    [Initiative.ELECTRIFICATION]: 'electrificationProject'
+  };
+
+  const projectTable = projectTableMap[initiative];
+
+  const sortDirection = options?.sortOrder === '1' ? 'asc' : 'desc';
+  const validSortFields = ['decisionDate', 'stage', 'state', 'statusLastChanged', 'submittedDate'];
+
+  // Default sorting
+  let orderBy: Record<string, 'asc' | 'desc'> = { submittedDate: 'desc' };
+
+  if (options?.sortField && validSortFields.includes(options.sortField)) {
+    orderBy = { [options.sortField]: sortDirection };
+  }
+
+  const whereClause = {
+    AND: [
+      {
+        activity: {
+          [projectTable]: {
+            some: {}
+          }
+        }
+      },
+      options.dateRange
+        ? {
+            OR: [
+              { submittedDate: { gte: options.dateRange[0], lte: options.dateRange[1] } },
+              { decisionDate: { gte: options.dateRange[0], lte: options.dateRange[1] } },
+              { statusLastChanged: { gte: options.dateRange[0], lte: options.dateRange[1] } }
+            ]
+          }
+        : {},
+      options?.permitTypeId ? { permitTypeId: parseInt(options.permitTypeId) } : {},
+      options?.sourceSystemKindId
+        ? {
+            permitTracking: {
+              some: {
+                sourceSystemKindId: parseInt(options.sourceSystemKindId)
+              }
+            }
+          }
+        : {},
+      options?.searchTag
+        ? {
+            OR: [
+              { activityId: { contains: options.searchTag, mode: 'insensitive' as const } },
+              { stage: { contains: options.searchTag, mode: 'insensitive' as const } },
+              { state: { contains: options.searchTag, mode: 'insensitive' as const } },
+              { permitType: { name: { contains: options.searchTag, mode: 'insensitive' as const } } },
+              { permitType: { businessDomain: { contains: options.searchTag, mode: 'insensitive' as const } } },
+              {
+                activity: {
+                  [projectTable]: {
+                    some: {
+                      OR: [
+                        { projectName: { contains: options.searchTag, mode: 'insensitive' as const } },
+                        { companyNameRegistered: { contains: options.searchTag, mode: 'insensitive' as const } },
+                        // Only include location fields for initiatives that have them (not ELECTRIFICATION)
+                        ...(initiative !== Initiative.ELECTRIFICATION
+                          ? [
+                              { streetAddress: { contains: options.searchTag, mode: 'insensitive' as const } },
+                              { locality: { contains: options.searchTag, mode: 'insensitive' as const } },
+                              { province: { contains: options.searchTag, mode: 'insensitive' as const } }
+                            ]
+                          : [])
+                      ]
+                    }
+                  }
+                }
+              },
+              {
+                permitTracking: {
+                  some: {
+                    trackingId: { contains: options.searchTag, mode: 'insensitive' as const }
+                  }
+                }
+              }
+            ]
+          }
+        : {}
+    ]
+  };
+
+  // Get total count (without pagination)
+  const totalRecords = await tx.permit.count({
+    where: whereClause
+  });
+
+  // Get paginated data
+  const permits = await tx.permit.findMany({
+    skip: options?.skip ? parseInt(options.skip) : 0,
+    take: options?.take ? parseInt(options.take) : 10,
+    where: whereClause,
+    orderBy: orderBy,
+    include: {
+      permitType: true,
+      permitTracking: {
+        include: {
+          sourceSystemKind: true
+        }
+      },
+      activity: {
+        include: {
+          [projectTable]: true
+        }
+      }
+    }
+  });
+
+  // Map the results to alias the project table as 'project'
+  const permitsWithProjectAlias = permits.map((permit) => {
+    const { [projectTable]: projectData, ...activityData } = permit.activity;
+
+    return {
+      ...permit,
+      activity: {
+        ...activityData,
+        project: projectData
+      }
+    };
+  });
+
+  return { permits: permitsWithProjectAlias, totalRecords };
 };
 
 /**
--- a/app/src/services/permit.ts
+++ b/app/src/services/permit.ts
@@ -221,12 +221,12 @@ export const searchPermitsPaginated = async (
             ]
           }
         : {},
-      options?.permitTypeId ? { permitTypeId: parseInt(options.permitTypeId) } : {},
+      options?.permitTypeId ? { permitTypeId: Number.parseInt(options.permitTypeId) } : {},
       options?.sourceSystemKindId
         ? {
             permitTracking: {
               some: {
-                sourceSystemKindId: parseInt(options.sourceSystemKindId)
+                sourceSystemKindId: Number.parseInt(options.sourceSystemKindId)
               }
             }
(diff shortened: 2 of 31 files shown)


slhurley:
- I don't think the File Number search is working...if I search by a specific ID it seems to more search on the Project number rather than the File number/tracking id.  It also brought up a false search by ID 5 for Transportation.
- Also the paging says Page 0 of 0 no matter how many on the list and you can't change the number on page
- It is not clear to me what date I am searching on with the Date Range search?
- I guess we can't do the automatic search like we have on Projects?  ie if I want to change the search I have to blank it out carefully first and then re-hit the Search button.  Is this a side effect of back-end pagination?

kyle1morel:
Doing any kind of date range or text search clears out the Authorization list, and it doesn't come back unless you hard refresh the page.

sanjaytkbabu:
> * I don't think the File Number search is working...if I search by a specific ID it seems to more search on the Project number rather than the File number/tracking id.  It also brought up a false search by ID 5 for Transportation.

Fixed!
> * Also the paging says Page 0 of 0 no matter how many on the list and you can't change the number on page

Fixed!
> * It is not clear to me what date I am searching on with the Date Range search?

Yes it can be confusing. But it was part of the a/c to search every date in the table. it's not searching just one column.
> * I guess we can't do the automatic search like we have on Projects?  ie if I want to change the search I have to blank it out carefully first and then re-hit the Search button.  Is this a side effect of back-end pagination?

Yes, the search as you type is a primevue table feature which we can't use in this case as the whole data isn't loaded to the table due to pagination.

For ease of clearing we could show 'clear all filters' as soon as they start typing, that way they don't have to backspace.
Also to search just hitting enter after typing would trigger the search, no need to hit the button.

If we absolutely have to do search as you type, we can do it the way we did for 'create user' on 'user management' page, it just comes with it's own overhead like managing duplicate api calls/late api response and stuff. We had  inconsistent search results on that page, but I think Kyle managed to eliminate most of the issues.

sanjaytkbabu:
> Doing any kind of date range or text search clears out the Authorization list, and it doesn't come back unless you hard refresh the page.

This should be fixed now

slhurley:
- Search by an exact ID is still not working all the time?
- Submitted Date is in a different format on the display than the other dates (I think Submitted format is the correct one)
- I wonder if the Date Range search field should be last so you don't have to skip over it to get to the regular Search box
- Back button and breadcrumb are taking you to Submissions rather than Authorizations
- Clear all filters isn't on the same line should it be...easier to spot?
- Could "No items found" be centered on the table?

minchang00:
- I also agree with Sharolyn moving the Date Range box to the end.

- Having the same problem with Sharolyn on the back-button. When I'm on the authorization status update page and use the back button, it goes back to the Project submission page rather than the Authorization list page.

- Keep getting the "Add user" modal when I use the breadcrumb to go to Project's Information tab. 

- Should we change the search box prompt to "Search ID #" after they select an ID so it shows that they can use the search box?

kyle1morel:
- Tracking ID drop down should be populated with all options if no Authorization type is selected, and subsequently filtered down based on selected Authorization. If this is not desired functionality then it should be disabled if there is no options available.
- Date Range selector is unclear. What dates are being filtered on?
- No way to clear a single filter. Instead of a `Clear all filters` option, recommend individual filters clears (see how https://shop.lululemon.com/en-ca/ does their filters for example)

kyle1morel:
> * Should we change the search box prompt to "Search ID #" after they select an ID so it shows that they can use the search box?

The search box is working as I would expect it to. It is an additional filter to string match a subset of your current filtered data. It should not be relegated to searching on one distinct column.

sanjaytkbabu:
> * Search by an exact ID is still not working all the time?

Should be working.
> * Submitted Date is in a different format on the display than the other dates (I think Submitted format is the correct one)

Fixed!
> * I wonder if the Date Range search field should be last so you don't have to skip over it to get to the regular Search box

Moved to the end as per discussions with Min. Apparently this is used less often.
> * Back button and breadcrumb are taking you to Submissions rather than Authorizations

Back button fixed! 
Breadcrumb navigation always takes us to the 1st one(Projects), that's how it works after opening an enquiry and navigating back to submissions using the breadcrumb. If this feature is needed, we'd need to look into it and implement it globally.
> * Clear all filters isn't on the same line should it be...easier to spot?

Left it as it is as per discussions with Min
> * Could "No items found" be centered on the table?

As discussed, it is centred in its own scrollable container

sanjaytkbabu:
> * I also agree with Sharolyn moving the Date Range box to the end.

Done!
> * Having the same problem with Sharolyn on the back-button. When I'm on the authorization status update page and use the back button, it goes back to the Project submission page rather than the Authorization list page.

Done!
> * Keep getting the "Add user" modal when I use the breadcrumb to go to Project's Information tab.

Fixed!
> * Should we change the search box prompt to "Search ID #" after they select an ID so it shows that they can use the search box?

Leaving it as it is as per Kyle's comment.

sanjaytkbabu:
> * Tracking ID drop down should be populated with all options if no Authorization type is selected, and subsequently filtered down based on selected Authorization. If this is not desired functionality then it should be disabled if there is no options available.

Tracking ID dropdown disabled
> * Date Range selector is unclear. What dates are being filtered on?

Leaving it as it is as per discussion with Min
> * No way to clear a single filter. Instead of a `Clear all filters` option, recommend individual filters clears (see how https://shop.lululemon.com/en-ca/ does their filters for example)

Leaving it as it is as per discussion with Min

slhurley:
- Quite often when first entering the Authorization tab, the dropdown for Authorization type says "No available options".  Seems to be on first entry.
- There seems to be an unusual amount of extra white space between columns in the last 4 columns on the list?

sanjaytkbabu:
> * Quite often when first entering the Authorization tab, the dropdown for Authorization type says "No available options".  Seems to be on first entry.

fixed!
> * There seems to be an unusual amount of extra white space between columns in the last 4 columns on the list?

updated!

app/src/controllers/permit.ts
```diff
@@ -113,6 +121,34 @@ export const sendPermitUpdateEmail = async (params: PermitUpdateEmailParams) =>
...
+      const options: SearchPermitsOptions = {
+        ...req.query
+      };
+
+      // Validate it's not PCNS
+      if (req.currentContext.initiative === Initiative.PCNS) {
+        throw new Problem(400, { detail: 'Invalid initiative' });
+      }
+
+      return await searchPermitsPaginated(tx, req.currentContext.initiative!, options);
```
sanjaytkbabu: response to sonar cloud's comment.
This is unfortunately necessary
'Argument of type 'Initiative.ELECTRIFICATION | Initiative.HOUSING | undefined' is not assignable to parameter of type 'Initiative.ELECTRIFICATION | Initiative.HOUSING'.
  Type 'undefined' is not assignable to type 'Initiative.ELECTRIFICATION | Initiative.HOUSING'

app/src/services/permit.ts
```diff
@@ -166,6 +173,147 @@ export const searchPermits = async (tx: PrismaTransactionClient, params: PermitS
...
+              { permitType: { businessDomain: { contains: options.searchTag, mode: 'insensitive' as const } } },
+              {
+                activity: {
+                  [projectTable]: {
+                    some: {
+                      OR: [
+                        { projectName: { contains: options.searchTag, mode: 'insensitive' as const } },
+                        { companyNameRegistered: { contains: options.searchTag, mode: 'insensitive' as const } },
+                        // Only include location fields for initiatives that have them (not ELECTRIFICATION)
+                        ...(initiative !== Initiative.ELECTRIFICATION
```
sanjaytkbabu: response to sonar cloud's comment.
I'm not sure if there is another way to conditionally include location column.

app/src/controllers/permit.ts
```diff
@@ -113,6 +121,34 @@ export const sendPermitUpdateEmail = async (params: PermitUpdateEmailParams) =>
...
+ */
+export const searchPermitsController = async (
+  req: Request<never, never, never, SearchPermitsOptions>,
+  res: Response
+) => {
+  const response = await transactionWrapper<{ permits: Permit[]; totalRecords: number }>(
+    async (tx: PrismaTransactionClient) => {
+      const options: SearchPermitsOptions = {
+        ...req.query
+      };
```
kyle1morel: No need to redeclare the query. Can pass `req.query` directly to the service call.

app/src/types/stuff.d.ts
```diff
@@ -346,6 +346,17 @@ export type DateTimeStrings = SplitDatetimeBase<string>;
...
+export interface SearchPermitsOptions extends Partial<IStamps> {
+  dateRange?: [Date, Date];
+  permitTypeId?: string;
+  searchTag?: string;
+  skip?: string;
+  sortField?: string;
+  sortOrder?: string;
+  sourceSystemKindId?: string;
+  take?: string;
+}
```
kyle1morel, line 415: `skip`, `take`, `sortField`, and `sortOrder` should probably be in their own interface, and have this extend it, as they will be used across all other pagination queries in the future.

Also unsure why this is extending `IStamps`.

app/src/validators/permit.ts
```diff
@@ -50,7 +50,22 @@ const schema = {
...
       activityId: Joi.string().min(8).max(8).allow(null),
-      includeNotes: Joi.boolean().allow(null)
+      includeNotes: Joi.boolean().allow(null),
+      pageNo: Joi.number().integer().min(1).allow(null),
+      permitTracking: permitTrackingSchema.allow(null),
+      permitType: permitTypeSchema.allow(null)
+    })
+  },
+  searchPermits: {
+    query: Joi.object({
```
kyle1morel, line 58: `skip`, `take`, `sortField`, and `sortOrder` should be in their own validator and be concatenated onto this one.

app/src/validators/permit.ts
```diff
@@ -50,7 +50,22 @@ const schema = {
   listPermits: {
     query: Joi.object({
       activityId: Joi.string().min(8).max(8).allow(null),
-      includeNotes: Joi.boolean().allow(null)
+      includeNotes: Joi.boolean().allow(null),
+      pageNo: Joi.number().integer().min(1).allow(null),
+      permitTracking: permitTrackingSchema.allow(null),
+      permitType: permitTypeSchema.allow(null)
```
kyle1morel: Mistakenly comitted?

app/src/validators/permit.ts
```diff
@@ -50,7 +50,22 @@ const schema = {
...
+    })
+  },
+  searchPermits: {
+    query: Joi.object({
+      dateRange: Joi.array().items(Joi.string()).length(2).allow(null),
+      permitTypeId: Joi.string().allow(null),
+      searchTag: Joi.string().allow(null),
+      skip: Joi.string().allow(null),
+      sortField: Joi.string().allow(null),
+      sortOrder: Joi.string().allow(null),
```
kyle1morel: We will want to specify a valid range for order, typically [-1, 0, 1].

app/src/services/permit.ts
```diff
@@ -166,6 +173,147 @@ export const searchPermits = async (tx: PrismaTransactionClient, params: PermitS
...
+): Promise<{ permits: Permit[]; totalRecords: number }> => {
+  // Determine project table based on initiative, exclude PCNS
+  const projectTableMap: Record<Exclude<Initiative, Initiative.PCNS>, string> = {
+    [Initiative.HOUSING]: 'housingProject',
+    [Initiative.ELECTRIFICATION]: 'electrificationProject'
+  };
+
+  const projectTable = projectTableMap[initiative];
+
+  const sortDirection = options?.sortOrder === '1' ? 'asc' : 'desc';
```
kyle1morel: Need to accept -1, 0, 1.
0 being default db ordering

app/src/services/permit.ts
```diff
@@ -166,6 +173,147 @@ export const searchPermits = async (tx: PrismaTransactionClient, params: PermitS
...
+  const validSortFields = ['decisionDate', 'stage', 'state', 'statusLastChanged', 'submittedDate'];
+
+  // Default sorting
+  let orderBy: Record<string, 'asc' | 'desc'> = { submittedDate: 'desc' };
```
kyle1morel: Default sort order is a front end concern. If the API is not provided anything it should return default db ordering.

frontend/src/components/authorization/AuthorizationListNavigator.vue
```diff
@@ -0,0 +1,381 @@
...
+import { Column, DataTable, DatePicker, IconField, InputIcon, InputText, Select } from '@/lib/primevue';
+import { useAppStore, useCodeStore } from '@/store';
+import { permitService, sourceSystemKindService } from '@/services';
+import { Initiative, RouteName } from '@/utils/enums/application';
+import { formatDateOnly } from '@/utils/formatters';
+import { generalErrorHandler, toNumber } from '@/utils/utils';
+
+import type { Ref } from 'vue';
+import type { Permit, Pagination, PermitType, PermitTracking, SourceSystemKind } from '@/types';
+import type { DataTableSortEvent } from 'primevue/datatable';
```
kyle1morel, line 15: alpha

frontend/src/components/authorization/AuthorizationListNavigator.vue
```diff
@@ -0,0 +1,381 @@
...
+  rows: 10,
+  order: -1,
+  field: 'submittedDate',
+  page: 0
+});
+
+// Interfaces
+interface InitiativeState {
+  projectAuthorizationRouteName: RouteName;
+}
```
kyle1morel: You've added this in the middle of `// State`

frontend/src/types/SearchPermitsOptions.ts
```diff
@@ -0,0 +1,12 @@
...
+export type SearchPermitsOptions = {
+  dateRange?: [Date, Date];
+  permitTypeId?: number;
+  searchTag?: string;
+  skip?: number;
+  sortField?: string;
+  sortOrder?: number;
+  sourceSystemKindId?: number;
+  take?: number;
+} & Partial<IStamps>;
```
kyle1morel: `skip`, `take`, `sortField`, and `sortOrder` should probably be in their own interface, and have this extend it, as they will be used across all other pagination queries in the future.

Also unsure why this is extending `IStamps`.

app/src/controllers/permit.ts
qhanson55: You can add `// nosonar` to the line causing the warning.

app/src/services/permit.ts
qhanson55: Flip the negation. 
```ts
...(initiative === Initiative.ELECTRIFICATION
  ? []
  : [
      { streetAddress: { contains: options.searchTag, mode: 'insensitive' as const } },
      { locality: { contains: options.searchTag, mode: 'insensitive' as const } },
      { province: { contains: options.searchTag, mode: 'insensitive' as const } }
    ])
```

frontend/src/components/submission/SubmissionsNavigator.vue
```diff
@@ -2,9 +2,9 @@
 import { addDays, isPast, isToday, isWithinInterval, startOfToday } from 'date-fns';
 import { storeToRefs } from 'pinia';
 import { inject, onBeforeMount, ref, watch, watchEffect } from 'vue';
-import { useRoute, useRouter } from 'vue-router';
 
 import { Escalation } from '@/components/common/icons';
+import AuthorizationListNavigator from '../authorization/AuthorizationListNavigator.vue';
```
qhanson55, line 6: import order

frontend/src/components/authorization/AuthorizationListNavigator.vue
```diff
@@ -0,0 +1,381 @@
...
+            loading.value = false;
+          });
+      } catch (e) {
+        generalErrorHandler(e);
+      }
+    });
+  }, 500);
+}
+
+// async function searchPermits() {
```
qhanson55: remove

frontend/src/components/authorization/AuthorizationListNavigator.vue
```diff
@@ -0,0 +1,381 @@
+<script setup lang="ts">
+import { Mutex } from 'async-mutex';
+import { storeToRefs } from 'pinia';
+import { onBeforeMount, ref } from 'vue';
+import { useI18n } from 'vue-i18n';
+import { useRoute } from 'vue-router';
+
+import { Column, DataTable, DatePicker, IconField, InputIcon, InputText, Select } from '@/lib/primevue';
+import { useAppStore, useCodeStore } from '@/store';
+import { permitService, sourceSystemKindService } from '@/services';
```
qhanson55, line 8: alpha

app/src/validators/permit.ts
sanjaytkbabu: yeah, thanks for catching!

app/src/validators/paginationOptions.ts
```diff
@@ -0,0 +1,8 @@
+import Joi from 'joi';
+
+export const paginationOptions = {
+  skip: Joi.string().allow(null),
+  sortField: Joi.string().allow(null),
+  sortOrder: Joi.string().allow(null).valid('-1', '0', '1'),
+  take: Joi.string().allow(null)
+};
```
kyle1morel: Should be a valid Joi `object`.

app/src/validators/permit.ts
```diff
@@ -53,16 +54,27 @@ const schema = {
       includeNotes: Joi.boolean().allow(null)
     })
   },
+  searchPermits: {
+    query: Joi.object({
+      dateRange: Joi.array().items(Joi.string()).length(2).allow(null),
+      permitTypeId: Joi.string().allow(null),
+      searchTag: Joi.string().allow(null),
+      sourceSystemKindId: Joi.string().allow(null),
+      ...paginationOptions
```
kyle1morel: Use the `Joi.concat` method to extend this with a valid `Joi.object`

frontend/src/components/authorization/AuthorizationListNavigator.vue
```diff
@@ -0,0 +1,362 @@
...
+  projectAuthorizationRouteName: RouteName.INT_ELECTRIFICATION_PROJECT_AUTHORIZATION
+};
+
+const GENERAL_INITIATIVE_STATE: InitiativeState = {
+  projectAuthorizationRouteName: RouteName.INT_GENERAL_PROJECT_AUTHORIZATION
+};
+
+const HOUSING_INITIATIVE_STATE: InitiativeState = {
+  projectAuthorizationRouteName: RouteName.INT_HOUSING_PROJECT_AUTHORIZATION
+};
```
kyle1morel: State should be provided at the view level, and injected into components.

frontend/src/types/SearchPermitsOptions.ts
```diff
@@ -0,0 +1,8 @@
+import type { IPaginationOptions } from '@/interfaces';
+
+export type SearchPermitsOptions = {
+  dateRange?: [Date, Date];
+  permitTypeId?: number;
+  searchTag?: string;
+  sourceSystemKindId?: number;
+} & Partial<IPaginationOptions>;
```
kyle1morel: Style: Prefer `interface`.

Unnecessary `Partial`. All keys are already optional.

frontend/src/store/projectStore.ts
```diff
@@ -110,7 +110,9 @@ export const useProjectStore = defineStore('project', () => {
       () =>
         // Look in activityContacts first, fall back to project state
         state.activityContacts.value.find((ac) => ac.role === ActivityContactRole.PRIMARY)?.contact ??
-        state.project.value?.activity?.activityContact?.find((ac) => ac.role === ActivityContactRole.PRIMARY)?.contact
+        state.project.value?.activity?.activityContact?.find(
+          (ac: ActivityContact) => ac.role === ActivityContactRole.PRIMARY
+        )?.contact
```
kyle1morel: Style: Type should be auto derived. But if adding an explicit type, why only update this line and not the one above?

frontend/src/services/permitService.ts
```diff
@@ -40,6 +40,14 @@ export default {
     return appAxios().get(`${useAppStore().getInitiative.toLowerCase()}/${PATH}`, { params: options });
   },
 
+  /**
+   * @function searchPermits
+   * @returns {Promise} An axios response
+   */
+  async searchPermits(options?: SearchPermitsOptions) {
```
kyle1morel: Function should have an explicit return type. Possibly `Promise<AxiosResponse<Permit[]>>`

frontend/src/components/authorization/AuthorizationListNavigator.vue
```diff
@@ -0,0 +1,362 @@
...
+const initiativeState: Ref<InitiativeState> = ref(HOUSING_INITIATIVE_STATE);
+const dateRange: Ref<[Date, Date] | undefined> = ref(undefined);
+const selection: Ref<Permit | undefined> = ref(undefined);
+const permits: Ref<Permit[]> = ref([]);
+const permitTypes: Ref<PermitType[]> = ref([]);
+const sourceSystemKinds: Ref<SourceSystemKind[]> = ref([]);
+const authorizationType: Ref<PermitType | undefined> = ref(undefined);
+const authorizationTracking: Ref<PermitTracking | undefined> = ref(undefined);
+const searchTag: Ref<string | undefined> = ref(undefined);
+const totalRecords: Ref<number> = ref(0);
```
kyle1morel, line 46: Style: alpha

frontend/src/components/authorization/AuthorizationListNavigator.vue
```diff
@@ -0,0 +1,362 @@
...
+      break;
+    case Initiative.GENERAL:
+      initiativeState.value = GENERAL_INITIATIVE_STATE;
+      break;
+    case Initiative.HOUSING:
+      initiativeState.value = HOUSING_INITIATIVE_STATE;
+      break;
+    default:
+      generalErrorHandler(t('views.initiativeStateError'));
+  }
```
kyle1morel: Should be removed in favour of injections

frontend/src/components/authorization/AuthorizationListNavigator.vue
```diff
@@ -0,0 +1,362 @@
...
+  pagination.value.field = typeof event.sortField === 'string' ? event.sortField : '';
+  pagination.value.order = event.sortOrder ?? 1;
+  pagination.value.page = 0;
+  searchPermits();
+}
+
+const searchMutex = new Mutex();
+let timeoutId: ReturnType<typeof setTimeout>;
+
+// Actions
```
kyle1morel: Style: Misplaced header comment

frontend/src/components/authorization/AuthorizationListNavigator.vue
```diff
@@ -0,0 +1,362 @@
...
+// read from query params if tab is set to enquiry otherwise use default values
+if (route.query.tab === AUTHORIZATION_TAB_INDEX.toString()) {
+  pagination.value.rows = toNumber(route.query.rows as string) ?? 10;
+  pagination.value.order = toNumber(route.query.order as string) ?? -1;
+  pagination.value.field = (route.query.field as string) ?? 'submittedAt';
+  pagination.value.page = toNumber(route.query.page as string) ?? 0;
+}
+
+// Actions
+onBeforeMount(async () => {
```
kyle1morel: Style: Lifecycle functions should appear at the bottom of actions

frontend/src/components/authorization/AuthorizationListNavigator.vue
```diff
@@ -0,0 +1,362 @@
...
+        (e) => {
+          pagination.page = e.page;
+          pagination.rows = e.rows;
+          searchPermits();
+        }
+      "
+    >
+      <template #empty>
+        <div class="flex justify-center">
+          <h3>No items found.</h3>
```
kyle1morel: i18n

frontend/src/components/authorization/AuthorizationListNavigator.vue
```diff
@@ -0,0 +1,362 @@
...
+        </div>
+      </template>
+      <template #header>
+        <div class="flex justify-between mb-3">
+          <div class="grid grid-cols-[2fr_1.5fr_1fr] gap-3">
+            <Select
+              v-model="authorizationType"
+              name="authorizationType"
+              :label="t('authorization.common.authorization')"
+              placeholder="Authorization type"
```
kyle1morel: i18n

frontend/src/components/authorization/AuthorizationListNavigator.vue
```diff
@@ -0,0 +1,362 @@
...
+                :placeholder="t('authorization.common.search')"
+                @update:model-value="searchPermits"
+              />
+            </IconField>
+          </div>
+        </div>
+      </template>
+      <Column
+        field="authorizationType"
+        header="Authorization Type"
```
kyle1morel: i18n

frontend/src/components/authorization/AuthorizationListNavigator.vue
```diff
@@ -0,0 +1,362 @@
...
+}
+
+:deep(.p-datatable-header) {
+  padding-right: 0;
+  padding-left: 0;
+}
+
+:deep(.p-datatable-loading-icon) {
+  color: var(--p-bcblue-800);
+}
```
kyle1morel: Consider making this a global style change

frontend/src/assets/main.scss
```diff
@@ -191,6 +191,9 @@ label {
 
 /* datatable */
 .p-datatable {
+  .p-datatable-loading-icon {
+  color: var(--p-bcblue-800);
+}
```
kyle1morel: Style: formatting

kyle1morel (approved):

kyle1morel (approved):

PR #488, opened 2026-06-10
chore: migrate Jest to Vitest and improve code test coverage

qhanson55:
# Description

[PADS-842](https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-842)

Migrates the backend test suite from Jest to Vitest, including config, mocks, and per-file syntax updates. While in there, adds unit tests to fill in coverage gaps across controllers (`code`, `sourceSystemKind`, `sso`, `map`, `yars`, `peach`, plus `user.getUsersWithGroups`), middleware (`authorization`, `requireSomeAuth`, `requireSomeGroup`, `validation`), and the `peach` service.

Vitest gives us a faster runner with native ESM/TS support and aligns the backend tooling with the frontend. The added tests raise backend coverage to ~93% statements / 79% branches and bring most of the touched files to 100%.

## Types of changes

Chore/Tests

## Checklist


kyle1morel (approved):

app/vitest.config.ts
```diff
@@ -0,0 +1,27 @@
+import { coverageConfigDefaults, defineConfig } from 'vitest/config';
+
+export default defineConfig({
+  test: {
+    coverage: {
+      provider: 'istanbul',
```
jujaga: I can't remember off the top of my head, but is the istanbul provider required for the pipe to work? There might have been something in the old Jest coverage that needed it back in the day, but Vitest should be able to work natively IIRC. You may be able to get away with `@vitest/coverage-v8` which is better.

app/vitest.config.ts
```diff
@@ -0,0 +1,27 @@
...
+      exclude: [
+        'src/**/*.d.ts',
+        'src/db/{extensions,manual-migrations,migrations,seeds}/**',
+        'src/db/codes/generate.ts',
+        'src/db/dataConnection.ts',
+        'src/db/stamps.ts',
+        'src/db/utils/transactionWrapper.ts',
+        'src/db/utils/yars.ts',
+        'src/interfaces/**',
+        'src/routes/utils.ts',
```
jujaga: I would be careful about adding specific files to the exclude list. While I can see they are mostly framework related tools for the most part, this can easily lead to test skewing down the line or outright missed tests/strange behaviour that might not be easy to catch just because of exclusion oversight. I'd recommend taking another look and seeing if they really should be skipped, even though it may drop the reported percentage in coverage.

sanjaytkbabu (approved):

app/src/db/utils/utils.ts
```diff
@@ -196,3 +196,17 @@ export async function dropAuditLogTrigger(knex: Knex, schema: string, table: str
 export async function dropUpdatedAtTrigger(knex: Knex, schema: string, table: string): Promise<void> {
   await knex.schema.raw(`DROP TRIGGER IF EXISTS before_update_${table}_trigger ON ${schema}.${table}`);
 }
+
+/**
+ * Add standard timestamp and user tracking columns to a table
+ * @param knex Knex instance
+ * @param table Table builder
+ */
+export function createStamps(knex: Knex, table: Knex.CreateTableBuilder) {
```
kyle1morel: Would probably call it `withTimestamps` or `withAuditFields` or something. `createStamps` to me implies only `created_by` and `created_at`. The front end now refers to them as `AuditFields`.

jujaga (approved):

kyle1morel (approved):

PR #498, opened 2026-06-17
refactor: rewrite filterActivityResponseByScope middleware

sanjaytkbabu:
# Description

Rewrites the `filterActivityResponseByScope` middleware.
- Removed as middleware
- It now exists as a function that must be used in the controller layer
- Automatically falls back to `activityId` and attempts a data lookup if no `activityContacts` are in the response. This allows us to not force `activityContact` includes into prisma so we can do filtering on the response object.
- Applies this new function to all locations that were using the middleware

[PADS-840](https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-840)

## Checklist


app/src/types/models.ts
```diff
@@ -8,7 +8,11 @@ export type AccessRequestPatch = Omit<Prisma.access_requestUpdateInput, 'accessR
 
 const activityBase = Prisma.validator<Prisma.activityDefaultArgs>()({});
 export type ActivityBase = Prisma.activityGetPayload<typeof activityBase>;
-export type Activity = ActivityBase & { activityContact?: ActivityContact[]; initiative?: Initiative };
+export type Activity = ActivityBase & {
+  activityContact?: ActivityContact[];
+  initiative?: Initiative;
+  project?: ElectrificationProject | GeneralProject | HousingProject | null;
```
kyle1morel: This should not be done. Projects are not a relation of an Activity. It also causes a cyclical loop.

frontend/src/types/api/resources.ts
```diff
@@ -373,13 +373,20 @@ interface PermitBase extends AuditFields {
...
 }
 
 export type Permit = PermitBase & Partial<PermitRelations>;
 
+export interface PermitWithActivityProject extends Permit {
+  activity: Activity & {
+    project: ElectrificationProject | GeneralProject | HousingProject | null;
+  };
+}
+
```
kyle1morel: This is not a first class citizen, so it doesn't belong in resources. It is a response type. Can probably be put directly into `SearchPermitsResponse` as it isn't used elsewhere. 

`project` should also be a `Nullable<Project>`

app/src/types/stuff.d.ts
```diff
@@ -393,6 +394,12 @@ export interface PermitUpdateEmailParams {
   emailTemplate: EmailTemplate;
 }
 
+export interface PermitWithActivityProject extends Permit {
+  activity: Activity & {
+    project: ElectrificationProject | GeneralProject | HousingProject | null;
```
kyle1morel: `project` should be `Project | null`. 
Or better yet, bring in the type utils that we have in the frontend and use `Nullable<Project>`

qhanson55 (approved):

PR #534, opened 2026-07-23
test: add shared frontend test utilities and testing standards

kyle1morel:
# Description

Introduces a shared set of frontend unit-test utilities (`mountComponent`, `mountWithFormContext`, `mockRouter`, `mockAuthNStore`) and a `TESTING-STANDARDS.md` doc codifying how specs should be structured going forward, then applies the new patterns to a batch of existing specs to clean up incorrect or noisy test setups.

- Added `frontend/tests/mountComponent.ts` and `frontend/tests/mountWithFormContext.ts` -- shared mounting helpers that centralize Pinia/PrimeVue/i18n/stub setup so individual specs stop hand-rolling `global` blocks.
- Added `frontend/tests/mockRouter.ts` -- a shared `mockRouter`/`resetMockRouter()` for specs asserting on `vue-router` navigation.
- Added `frontend/tests/mockAuthNStore.ts` -- mocks `useAuthNStore` wholesale to work around `oidc-client-ts`'s `SessionMonitor`, which otherwise logs `"[UserManager] getUser: user not found in storage"` in any spec that exercises the real authn store (the store's Pinia setup body always constructs a real `AuthService`/`UserManager`, even under `createTestingPinia`).
- Added `frontend/tests/TESTING-STANDARDS.md` documenting the mounting utilities, router/authn mocking conventions, the local-mount-factory pattern, file layout, and several Pinia/vue-test-utils gotchas discovered while doing this cleanup.
- Reworked ~25 existing spec files (`ProjectIntakeForm`, `ProjectListNavigator`, `ContactCardNavForm`, `NaturalDisasterCard`, `EnquiryIntakeForm`, `HeaderMenu`, etc.) to use the new utilities, fix warnings from improper test setups, and in a couple of cases fix latent bugs (e.g. `HeaderMenu.spec.ts` was seeding Pinia state under the wrong store key, so it never actually exercised the authenticated code path).
- One production fix: `ProjectIntakeForm.vue`'s `onBeforeMount` callback no longer needs to be `async` (no `await` in its body), which was flagged while updating its spec.

https://apps.nrs.gov.bc.ca/int/jira/browse/PADS-759

## Types of changes

Documentation (non-breaking change with enhancements to documentation)

## Checklist

### DIFF
--- a/frontend/tests/TESTING-STANDARDS.md
+++ b/frontend/tests/TESTING-STANDARDS.md
@@ -1,0 +1,296 @@
+# Frontend Test Standards
+
+This document is the standard for how test files in this codebase should
+be structured going forward. New and refactored specs should follow it.
+It isn't exhaustive -- add to it as new patterns get established.
+
+## Mounting utilities
+
+Two shared entry points live in `tests/`. Pick based on what the
+component under test actually needs:
+
+| Utility | Use when the component... |
+|---|---|
+| `mountComponent` | Establishes its own `<Form>` context internally, or needs no vee-validate form context at all -- just Pinia / provide / stubs. This is the default. |
+| `mountWithFormContext` | **Assumes an ancestor** already provides vee-validate's form context (a field/section component that doesn't render its own `<Form>`). |
+
+`mountWithFormContext` mounts the component inside a synthetic harness
+that calls `useForm()` on its behalf. Reach for it only when the
+component genuinely has no other way to get a form context in the test
+-- using it on a component that owns its own `<Form>` creates a nested,
+shadowed context instead of an error, which is a quiet bug to track down
+later. When in doubt, start with `mountComponent`; only switch to
+`mountWithFormContext` if the component throws on a missing form
+context (e.g. `useIsFormDirty`/`useFormValues`/`useField` with no
+provider).
+
+## Router mocking
+
+`tests/mockRouter.ts` exports a shared `mockRouter` (`{ push, replace }`)
+and `resetMockRouter()`. It's deliberately separate from the mounting
+utilities above -- routing is an unrelated concern, not something every
+mounted component needs. Use it in any spec that needs to assert on
+navigation:
+
+```ts
+import { mockRouter, resetMockRouter } from '../utils/mockRouter';
+
+// vi.mock is hoisted, so it can't reference `mockRouter` directly from the
+// shared util at declaration time -- but it CAN return the same object
+// reference at call time, which is all that matters for spies.
+vi.mock('vue-router', () => ({
+  useRouter: () => mockRouter,
+  useRoute: () => ({ params: {}, query: {} })
+}));
+
+beforeEach(() => {
+  resetMockRouter();
+});
+
+// later, in a test:
+expect(mockRouter.replace).toHaveBeenCalledWith({ name: RouteName.EXT_GENERAL });
+```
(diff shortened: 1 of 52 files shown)


frontend/tests/TESTING-STANDARDS.md
```diff
@@ -0,0 +1,251 @@
...
+
+beforeEach(() => {
+  resetMockRouter();
+});
+
+// later, in a test:
+expect(mockRouter.replace).toHaveBeenCalledWith({ name: RouteName.EXT_GENERAL });
+```
+
+If a spec needs `useRoute()` to return different query params per test
```
qhanson55, line 54: Consider giving example like above

frontend/tests/TESTING-STANDARDS.md
```diff
@@ -0,0 +1,251 @@
...
+auto-stubbed -- the setup() body itself still runs for real, which
+unconditionally constructs a real `AuthService` / `oidc-client-ts`
+`UserManager`. With `monitorSession` on by default, that spins up a
+`SessionMonitor` whose constructor fires an unawaited `getUser()`, logging
+`"[UserManager] getUser: user not found in storage"` to the console in
+every spec that touches the real store -- there's no `piniaState` to seed
+around this, since it happens before any state is read.
+
+Because of that, specs depending on `useAuthNStore` are a deliberate
+exception to the "seed real Pinia state, don't mock the store" rule below:
```
qhanson55: Consider linking to the mentioned rule/section.

frontend/tests/TESTING-STANDARDS.md
```diff
@@ -0,0 +1,251 @@
...
+use `tests/mockAuthNStore.ts` instead.
+
+```ts
+import { mockAuthNStore, resetMockAuthNStore } from '../../../mockAuthNStore';
+
+vi.mock('@/store/authnStore', () => ({
+  default: () => mockAuthNStore,
```
qhanson55, line 126: why `default` and `useAuthNStore`

frontend/tests/TESTING-STANDARDS.md
```diff
@@ -0,0 +1,251 @@
...
+comment banner (not a boxed/bordered one -- keep it light):
+
+```ts
+// Mocks
+
+vi.mock('vue-router', () => ({ ... }));
+
+// Fixtures
+
+const testFoo = { ... };
```
qhanson55, line 195: Do we want a file that holds reusable test data like the back end?

frontend/tests/TESTING-STANDARDS.md
kyle1morel: Stores have 2 exports, need to ensure both possible import paths are mocked
```ts
export const useAuthNStore = defineStore() {}
export default useAuthNStore;
```

frontend/tests/TESTING-STANDARDS.md
kyle1morel: A lot of back end tests genuinely share the same shape. The front end is dealing with shapes tailored to that components specific needs. I think having a dumping ground like our back end would be much harder to maintain here. If we do start to notice a lot of crossover as more tests get written out I'm not opposed to starting one though.

qhanson55 (approved):

jujaga (approved):

